PR #23, opened 2015-04-13
Fix usage of deprecated method for request host

arjenvanderende:
Fix deprecation warnings in the log when using the express handler.

```js
express deprecated req.host: Use req.hostname instead node_modules/raygun/lib/raygun.messageBuilder.js:114:26
```


jamiepenney:
Thanks for this

PR #30, opened 2015-06-09
Add onBeforeSend function, like raygun4js

lee-houghton:


jamiepenney:
Moved this work to #35

PR #31, opened 2015-06-09
Handle network errors when uploading to api.raygun.io

lee-houghton:
This avoids unhandled exceptions due to failed DNS lookups or certificate issues. I also took the liberty of extending the callback to allow receiving an error, if the callback takes more than one parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,13 @@ The [Express documentation](http://expressjs.com/guide/error-handling.html) says
 
 ## Documentation
 
+### Callbacks
+
+The callback should be a node-style callback: `function(err, response) { /*...*/ }`.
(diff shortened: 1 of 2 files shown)


lee-houghton:
Looks like I was working from an out-of-date version here, but there doesn't seem to be much of a difference. 

I don't know why, but DNS errors seem to be quite frequent for api.raygun.io. When DNS errors occur, they aren't handled, and Node's EventEmitter's special handling of the `error` event causes an unhandled exception to be raised. If you don't want the change of callback signature, at least adding the `request.on("error")` would be enough :)

fundead:
Apologies for the delay in response to this PR, and thanks for you work on it. This has now been merged and will go onto NPM shortly as v0.9.0.

PR #48, opened 2015-11-12
allow passing in a string as an error

fjakobs:
it happened in our team a few times that developers would pass a string instead of an error instance to Raygun and then the log was full of `NoMessage` errors. This is an easy fix for this. I'm sure others have run into this as well.


dwnz:
Awesome, thanks for submitting this!

dwnz:
Pushed to npm - version 0.8.3

fjakobs:
awesome. thanks also for re-publishing

PR #50, opened 2015-12-03
Handle request errors and log out to console

timjrobinson:
Request errors should be caught and logged out, otherwise the node process will crash if it can't connect to Raygun.


dwnz:
Looks good - thanks for submitting this!

dwnz:
Pushed to npm - version 0.8.3

PR #58, opened 2016-02-14
fixed useSSL option, and support for using proxies.

danielcondemarin:
Fixed useSSL option passed in the ```init() options```, which was always evaluating to truthy. Also added proper support for the use of http proxies, by adding the host header on the request set to 'api.raygun.io', and setting the full path to the /entries url, which is all required for being able to send the request through a proxy (http://stackoverflow.com/questions/3862813/how-can-i-use-an-http-proxy-with-node-js-http-client).

An example setup would look like:

```
var raygun = require('raygun');

var raygunClient = new raygun.Client().init({ 
   apiKey: 'xxxx',
   host: 'myproxy.somedomain.co.uk',
   useSSL : false,
   port : 3128
});
```

### DIFF
--- a/lib/raygun.js
+++ b/lib/raygun.js
@@ -22,7 +22,7 @@ var Raygun = function () {
         _filters = options.filters;
         _host = options.host;
         _port = options.port;
-        _useSSL = options.useSSL || true;
+        _useSSL = options.useSSL !== false;
         _onBeforeSend = options.onBeforeSend;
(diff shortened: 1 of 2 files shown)


danielcondemarin:
ping

fundead:
Firstly, apologies for the delay in response to this PR - it has now been merged and is going out now as v0.9.0. Thanks for your work on this one!

PR #59, opened 2016-03-25
treat custom errors as Error's

mvhenten:
When extending the Error object (many libs do this) raygun treats them as plain Object.
By replacing this check with an instanceof, error objects will be treated as such.

### DIFF
--- a/lib/raygun.messageBuilder.js
+++ b/lib/raygun.messageBuilder.js
@@ -50,8 +50,6 @@ var RaygunMessageBuilder = function (options) {
   };
 
   this.setErrorDetails = function (error) {
-    var errorType = Object.prototype.toString.call(error);
-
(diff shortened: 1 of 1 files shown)


lib/raygun.messageBuilder.js
```diff
@@ -52,7 +52,7 @@ var RaygunMessageBuilder = function (options) {
   this.setErrorDetails = function (error) {
     var errorType = Object.prototype.toString.call(error);
```
nonken: not used anymore?

mvhenten: idd tests fail due to this! how awesome. removing it.

fundead:
Just an update on this one, this has now been merged and will be available from NPM shortly as v0.9.0. Thanks for your work on this one!

PR #66, opened 2018-05-25
Inner error support

### LINKED ISSUE #64: Support for VError
VError(https://github.com/joyent/node-verror) supports nesting errors as `cause`, which is similar to innerException in C#

According to docs (https://raygun.com/docs/integrations/api) it appears to be fairly trivial to add support for it in raygun4node to get nice nested errors in Raygun.

If this sounds like a plausible feature, I am willing to create a PR for it.

Deadarius:
# In this PR:
- inner errors are included in message sent to raygun API
- new option `innerErrorFieldName`
- out of the box support for [VError](https://github.com/joyent/node-verror)
- no breaking changes

closes #64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.10.0
+- Add support for inner errors. Option `innerErrorFieldName` to specify a field or a function on the error object to use for retrieval of an inner error. Defaults to `cause` which is used in [VError](https://github.com/joyent/node-verror)
+
 ## 0.9.1
 
 - Add an option to report column number for each frame of the stack trace
--- a/README.md
+++ b/README.md
@@ -139,6 +139,10 @@ Release 0.3 previously had a setUser function that accepted a string or function
 
 Call setVersion(*string*) on a RaygunClient to set the version of the calling application. This is expected to be of the format x.x.x.x, where x is a positive integer. The version will be visible in the dashboard.
 
+### Inner Errors
+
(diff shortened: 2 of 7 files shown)


UberMouse:
Hey @Deadarius,

Cheers for this. I'll take a look when I can, not sure exactly when that will be but sometime in the next week or so.

UberMouse:
Hi @Deadarius,

I can't get this to work with VError, what version of Node are you using? The behavior I am experiencing is whenever `cause` is called an assertation fails in the cause function of the error not being an `Error`. This seems to stem from the fact the code is calling `VError.prototype.cause` when the recommended method is `VError.cause` since we don't want to depend on the `VError` library that is unfortunately not possible. 

`VError.prototype.cause` calls `VError.cause` with `this` as an argument and it appears to be passing in the global Node environment for me, not the `VError` error instance it is called in.

Any information on how you got it to function would be appreciated.

Deadarius:
@UberMouse my bad, I was initially using `jse_cause` internal field of `VError` but than figured it would be better to call `.cause`, turns out I forgot to rerun it in conjunction with VError. I will update the PR accordingly and ping you.

Deadarius:
Hi @UberMouse,

Updated a PR, now `.cause` function on `VError` will be called with the right `this`.
Also since after merge of this PR `VError` support will be included in documentation it felt logical to me to include `VError` into `devDependencies` and write a couple of tests making sure that it works.

Feel free to exclude `VError` tests commit if you think it's unreasonable.

Cheers,
Alex.

UberMouse:
Thanks @Deadarius, I'll take a look at this again next week when I have the time.

UberMouse:
Just fixed up the tests and merged it in @Deadarius. Thanks for this.

PR #84, opened 2020-04-10
Update express example

Widdershin:
Primarily just updating vulnerable dependencies, but I also swapped out `node-compass` for `node-sass-middleware` to save a Ruby dependency.

This PR also adds a `package-lock.json`, which pins exact dependency versions to help ensure it's easier to run this example in the future.


Widdershin:
Cheers @BenjaminHarding, thanks for the review :)

PR #85, opened 2020-04-10
Replace deprecated Buffer constructor with Buffer.from

Widdershin:
`new Buffer` has been deprecated and now emits a message. The particular
form of it we were using was not a vulnerability as far as I am aware,
due to passing a string value rather than a manual size, but it's still
recommended to update to `Buffer.from`.

https://nodejs.org/gl/docs/guides/buffer-constructor-deprecation/


Widdershin:
Thanks @MattByers 👍

PR #86, opened 2020-04-10
Typescript

Widdershin:
This PR overhauls the project to TypeScript strict mode.

The main rationale for this was easing the process of adding new features to this repo without compromising existing behaviour, and having a reliable set of TypeScript definitions for our users that are guaranteed to stay up to date.

As much as possible this pull request retains all the original behaviour of the module prior to refactor, even if that meant compromising on type safety for the time being.

This means that some of the annotations feel a little tortured, as there are a few places where we initialize a class and then call `init`, rather than fully constructing the object up front. This means that the type system has to reflect the possibility that the user never called `.init`, which is a real concern, but adds a lot of uncertainty to the types.

On the whole this feels worth it, especially for having accurate type definitions we can expose to users.

### DIFF
--- a/lib/raygun.messageBuilder.ts
+++ b/lib/raygun.messageBuilder.ts
@@ -26,6 +26,8 @@ import {
   BuiltError
 } from './types';
 
+type UserMessageData = RawUserData | string | undefined;
+
 const humanString = require('object-to-human-string');
 const packageDetails = require('../package.json');
 
@@ -181,8 +183,9 @@ export class RaygunMessageBuilder {
     return this;
   };
 
-  setUser(user: (() => RawUserData) | RawUserData) {
-    var userData: RawUserData;
+  setUser(user: (() => UserMessageData) | UserMessageData) {
+    var userData: UserMessageData;
+
     if (user instanceof Function) {
       userData = user();
     } else {
@@ -191,7 +194,7 @@ export class RaygunMessageBuilder {
 
     if (userData instanceof Object) {
         this.message.details.user = this.extractUserProperties(userData);
-    } else {
+    } else if (typeof userData === "string") {
         this.message.details.user = { 'identifier': userData };
     }
 
--- a/lib/raygun.messageBuilder.ts
+++ b/lib/raygun.messageBuilder.ts
@@ -8,12 +8,28 @@
 
 'use strict';
 
-var stackTrace = require('stack-trace');
-var os = require('os');
-var humanString = require('object-to-human-string');
-var packageDetails = require('../package.json');
+import os from 'os';
+import stackTrace from 'stack-trace';
 
-function filterKeys(obj, filters) {
+import {
+  MessageBuilderOptions,
+  StackFrame,
+  Message,
+  MessageBuilding,
+  IndexableError,
+  RawUserData,
+  UserDetails,
+  RequestParams,
+  Tag,
+  CustomData,
+  Environment,
+  BuiltError
+} from './types';
+
+const humanString = require('object-to-human-string');
+const packageDetails = require('../package.json');
+
+function filterKeys(obj: {[key: string]: any}, filters: string[]) {
   if (!obj || !filters || typeof obj !== 'object') {
     return obj;
   }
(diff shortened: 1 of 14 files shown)


Widdershin:
Node 4 and 6 are now deprecated, and that means that ecosystem support has dropped off for them. Tap can no longer run tests for those versions as a result.

I've updated the `.travis.yml` to stop testing those Node versions, since they're no longer receiving security updates, it doesn't seem worth the hassle to support them.

lib/raygun.ts
```diff
@@ -0,0 +1,242 @@
...
+    this._offlineStorageOptions = options.offlineStorageOptions;
+    this._isOffline = options.isOffline;
+    this._groupingKey = options.groupingKey;
+    this._tags = options.tags;
+    this._useHumanStringForObject =
+      options.useHumanStringForObject === undefined
+        ? true
+        : options.useHumanStringForObject;
+    this._reportColumnNumbers = options.reportColumnNumbers;
+    this._innerErrorFieldName = options.innerErrorFieldName; // VError function to retrieve inner error;
```
BenjaminHarding, line 88: Just to confirm, we no longer need the `|| 'cause'` assignment here?

lib/raygun.ts
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  send(
+    exception: Error | string,
+    customData: CustomData,
+    callback: (err: Error | null) => void,
+    request?: Request,
+    tags?: Tag[]
+  ): Message {
+    var mergedTags: Tag[] = [];
```
BenjaminHarding, line 150: Nit: Use `let`?

lib/raygun.ts
```diff
@@ -0,0 +1,242 @@
...
+      this.offlineStorage().save(transportMessage, callback);
+    } else {
+      raygunTransport.send(transportMessage);
+    }
+
+    return message;
+  }
+
+  expressHandler(err: Error, req: Request, res: Response, next: NextFunction) {
+    var customData;
```
BenjaminHarding, line 219: Nit: Use `let`?

lib/raygun.messageBuilder.ts
```diff
@@ -0,0 +1,248 @@
...
+        queryString: filterKeys(request.query, this._filters),
+        headers: filterKeys(request.headers, this._filters),
+        form: filterKeys(request.body, this._filters),
+      };
+    }
+    return this;
+  }
+
+  setUser(user: (() => UserMessageData) | UserMessageData) {
+    var userData: UserMessageData;
```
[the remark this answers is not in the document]
Widdershin: Fixed 👍

lib/raygun.ts
Widdershin: Fixed 👍

lib/raygun.ts
Widdershin: Fixed 👍

lib/raygun.ts
Widdershin: Ah I had moved this because at one point the message builder required `innerErrorFieldName` to be a string, and since this class uses the `init` pattern `this._innerErrorFieldName` might be possibly `undefined`. The message builder handles not being passed an `innerErrorFieldName` now, so I've undone this change.

lib/raygun.messageBuilder.ts
```diff
@@ -0,0 +1,248 @@
...
+
+export class RaygunMessageBuilder {
+  _filters: string[];
+  options: MessageBuilderOptions;
+  message: MessageBuilding;
+
+  constructor(options: MessageBuilderOptions = {}) {
+    options = options || {};
+    this.options = options;
+    this._filters = options.filters || [];
```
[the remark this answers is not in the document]
Widdershin: Since this is an internal class we can rely on the type checker to ensure that `options.filters` is an array, so we don't need to handle those cases at runtime anymore, so it should be okay :)

lib/raygun.offline.ts
```diff
@@ -0,0 +1,117 @@
...
+      err,
+      cacheContents
+    ) {
+      raygunTransport.send(JSON.parse(cacheContents));
+      fs.unlink(path.join(storage.cachePath, item), () => {});
+    });
+  }
+
+  init(offlineStorageOptions: OfflineStorageOptions | undefined) {
+    if (!offlineStorageOptions || !offlineStorageOptions.cachePath) {
```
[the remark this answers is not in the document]
Widdershin: Thanks :) full credit to the compiler for this bug fix, it's an error otherwise!

PR #87, opened 2020-04-17
Notify apm of exceptions

Widdershin:
This branch adds integration with `raygun-apm` that allows for linked exceptions in the Raygun app.

We now generate a UUIDv4 to serve as a correlation id, as pass that both to the crash reporting endpoint and to the apm module. This allows the generated trace to reference the linked exception correctly.

### DIFF
--- a/lib/raygun.ts
+++ b/lib/raygun.ts
@@ -34,8 +34,10 @@ import { v4 as uuidv4 } from "uuid";
 
 const debug = require("debug")("raygun");
 
-
-let apmBridge: undefined | null | {notify(e: string | Error, s: string): void} = undefined;
+let apmBridge:
+  | undefined
+  | null
+  | { notify(e: string | Error, s: string): void } = undefined;
 
 try {
   if (module.parent) {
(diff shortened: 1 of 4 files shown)


lib/raygun.ts
```diff
@@ -30,8 +30,23 @@ import { OfflineStorage } from "./raygun.offline";
...
+try {
+  if (module.parent) {
+    apmBridge = module.parent.require('raygun-apm/lib/src/crash_reporting');
+  } else {
+    apmBridge = require('raygun-apm/lib/src/crash_reporting');
+  }
+} catch (e) {
+  apmBridge = null;
+}
+
```
Widdershin, line 49: This looks a little complex, but in essence we need require the `raygun-apm` module from the parent's require scope to have access. If for any reason we aren't able to do that, linked exceptions will be disabled.

lib/raygun.ts
```diff
@@ -248,6 +269,8 @@ class Raygun {
           : message;
     }
 
+    message.details.correlationId = correlationId;
```
BenjaminHarding, line 272: Should the message always have a `correlationId` or only if the `apmBridge` has been used?

Widdershin: Good point, might as well save the bytes if the `apmBridge` isn't present :)

Widdershin:
Thanks @BenjaminHarding for the snappy review 😸

BenjaminHarding:
Anytime :D

PR #88, opened 2020-04-21
Tag unhandled exceptions from express middleware

Widdershin:
This PR has been opened against the `typescript` branch, to simplify the diff. Once the `typescript` branch is merged this PR should be updated to merge into master.

-----

Previously, this provider made no differentiation between handled
errors (where a user directly reports a caught error with .send) and an
unhandled error, where an indirect mechanism like the express error
handler is used to catch and report an error.

Since the express middleware is currently the only source of unhandled
errors, we can start by adding the "UnhandledException" tag to all
errors caught by the express middleware.

The bulk of this change adds a test to exercise the express middleware.
Other tests make use of `nock` to stop the tests from hitting the API,
but this is a tad inelegant for our case as we want to allow local
traffic, and ideally find out what the raygun API server would be
receiving.

The `makeClientWithMockServer` will create an instance of the Raygun
client, as well as an express server that mimics the crash entries
endpoints. This means we can use the client and any errors generated
will be submitted to our mock server, which we can then inspect
directly.

This is nice because it's the first test that actually verifies that the
data is submitted to the server in the expected format. Other tests
merely check that the message has been constructed correctly, but are
still vulnerable to transport failures.

### DIFF
--- a/lib/raygun.ts
+++ b/lib/raygun.ts
@@ -224,7 +224,9 @@ class Raygun {
       customData = this.expressCustomData;
     }
 
-    this.send(err, customData || {}, function () {}, req);
+    this.send(err, customData || {}, function () {}, req, [
+      "UnhandledException",
+    ]);
     next();
   }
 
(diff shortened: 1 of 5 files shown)


Widdershin:
Investigating the build failure, unsurprising that node 4 and 6 don't work, but would expect that node 8 does.

Widdershin:
Thanks, I'll retarget and merge once the typescript branch is merged.

PR #89, opened 2020-05-06
Support batched error reporting

Widdershin:
You can now pass {batch: true} when calling `.init` to enable the batch
transport.

When this option is enabled, errors will be placed in a queue and
processed at a regular interval.

This should cut down on transmission overhead if there are a large
number of errors in short succession.

You can further control the frequency at which batches are processed by
providing a `batchFrequency` option when initializing, which specifies
the interval between batches in milliseconds.

TODO:
- [x] Support offline mode
- [ ] ~~Documentation~~ will be done in a separate PR and merged when changes are released
- [x] More manual testing

### DIFF
--- a/lib/raygun.batch.ts
+++ b/lib/raygun.batch.ts
@@ -91,7 +91,7 @@ export class RaygunBatchTransport {
     const payload = `[${batch.join(",")}]`;
     const runAllCallbacks = <E, R>(err: E, response: R) => {
       const [seconds, nanoseconds] = process.hrtime(startTime);
-      const durationInMs = Math.round(seconds / 1000 + nanoseconds / 1e6);
+      const durationInMs = Math.round(seconds * 1000 + nanoseconds / 1e6);
       if (err) {
         debug(
           `batch transport - error sending batch (id=${batchId}, duration=${durationInMs}ms): ${err}`
--- a/lib/raygun.batch.ts
+++ b/lib/raygun.batch.ts
@@ -1,11 +1,13 @@
-import { Message, HTTPOptions } from "./types";
 import { send } from "./raygun.transport";
+import { startTimer } from './timer';
+import type { IncomingMessage } from "http";
+import { callVariadicCallback, Callback, Message, HTTPOptions } from "./types";
 
 const debug = require("debug")("raygun");
 
 export type MessageAndCallback = {
   serializedMessage: string;
-  callback: Function;
+  callback: Callback<IncomingMessage>;
 };
 
 const MAX_MESSAGES_IN_BATCH = 100;
@@ -24,7 +26,7 @@ export class RaygunBatchTransport {
     this.httpOptions = options.httpOptions;
   }
 
-  send(message: string, callback: Function) {
+  send(message: string, callback: Callback<IncomingMessage>) {
     this.messageQueue.push({ serializedMessage: message, callback });
   }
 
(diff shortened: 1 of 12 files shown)


lib/raygun.batch.ts
```diff
@@ -0,0 +1,125 @@
...
+  stopProcessing() {
+    if (this.intervalId) {
+      debug("batch transport - stopping");
+      clearInterval(this.intervalId);
+
+      this.intervalId = null;
+    }
+  }
+
+  private process() {
```
Widdershin, line 47: Upon reflection this could probably use unit tests, I'll add those.

lib/raygun.ts
```diff
@@ -11,44 +9,28 @@
...
-  offlineStorage?: OfflineStorage;
-  offlineStorageOptions?: OfflineStorageOptions;
-  isOffline?: boolean;
-  groupingKey?: Hook<string>;
-  tags?: Tag[];
-  useHumanStringForObject?: boolean;
-  reportColumnNumbers?: boolean;
-  innerErrorFieldName?: string;
-};
+const DEFAULT_BATCH_FREQUENCY = 1000; // ms
```
Widdershin, line 33: Not sure what a good default would be here. Perhaps a bit longer of a period, but needs benchmarking to see how much that impacts heavy load.

lib/raygun.offline.ts
```diff
@@ -30,18 +27,30 @@ export class OfflineStorage {
       err,
       cacheContents
     ) {
-      raygunTransport.send(JSON.parse(cacheContents));
+      if (!storage.transport) {
+        throw new Error("OfflineStorage was initialized without transport");
+      }
```
BenjaminHarding, line 32: Would it be better for this check to be included in the OfflineStorages `init` method?

BenjaminHarding: I suppose it shouldn't matter too much because this is more of an internal class

lib/raygun.ts
```diff
@@ -140,14 +141,70 @@ class Raygun {
     this._tags = tags;
   }
 
+  transport() {
```
BenjaminHarding, line 144: Is there a specific interface/type that can be specified between the transport types to keep the externally accessible props consistent?

lib/raygun.ts
```diff
@@ -140,14 +141,70 @@ class Raygun {
...
+          message,
+          callback: wrappedCallback,
+          batch: false,
+          http: {
+            host: client._host,
+            port: client._port,
+            useSSL: !!client._useSSL,
+            apiKey,
+          },
+        });
```
BenjaminHarding, line 195: Might not be worth it but this looks similar to the batch transports send method so it could be good to consolidate these?

lib/raygun.offline.ts
Widdershin: Good point! This error check only existed to please the compiler, since the transport might be undefined if we forgot to call init.

I've refactored this so that the transport is passed in via the constructor, so we don't need to worry about it being undefined inside of that class anymore.

lib/raygun.ts
Widdershin: Good catch, added an annotation :)

lib/raygun.ts
Widdershin: Agree, there's a few details here that would be better not to duplicate. I've refactored the timing and callback logic so they're shared and it feels nicer.

lib/raygun.batch.ts
Widdershin: Done

lib/raygun.ts
Widdershin: I think this will actually be okay to start with. It's easy to change both locally and in releases, so we can worry about it later.

PR #90, opened 2020-06-29
fix: Move express types to dependencies

ffflorian:
Since the express types are imported and used in [`raygun.ts`](https://github.com/MindscapeHQ/raygun4node/blob/master/lib/raygun.ts), they should be shipped with raygun. Otherwise the user needs to add the types manually or this error will appear:

```
node_modules/raygun/build/raygun.d.ts(2,54): error TS2307: Cannot find module 'express' or its corresponding type declarations.
```


Widdershin:
Thanks for the PR @ffflorian, merged and released as `v0.12.1`.

PR #96, opened 2020-11-18
[WIP] Add support for breadcrumbs

Widdershin:
Currently built on top of `AsyncLocalStorage`, which means that we can run a section of code with a particular scope for breadcrumbs, which allows us to process multiple requests in parallel and still attribute breadcrumbs correctly without a great deal of user intervention.

Note that `AsyncLocalStorage` is only available on Node `12.17+` and onwards. An implementation that provides this sort of behaviour for earlier Node versions would have to rely on either re-implementing `AsyncLocalStorage` by hand with async hooks, or using Domains, which is [not always a good idea](https://nodejs.org/en/docs/guides/domain-postmortem/).

I personally see this as an acceptable tradeoff, as it provides support for the latest version of all LTS builds, and as currently implemented it simply does nothing on earlier versions.

Todo:
 - [x] Populate className, methodName and lineNumber if possible
 - [x] Dedicated express middleware
 - [ ] User API for re-entering a breadcrumb scope (after a DB request for example)
 - [ ] Investigate re-using async hook patches from `raygun-apm`
 - [ ] Documentation
 - [ ] More comprehensive unit testing
 - [ ] Warn if Node version doesn't support `AsyncLocalStorage`

### DIFF
--- a/lib/breadcrumbs.ts
+++ b/lib/breadcrumbs.ts
@@ -1,5 +1,6 @@
 import type { AsyncLocalStorage } from "async_hooks";
 import type { Breadcrumb, InternalBreadcrumb } from "./types";
+import path from "path";
 const debug = require("debug")("raygun").bind(null, "[breadcrumbs]");
 
 let asyncLocalStorage: AsyncLocalStorage<InternalBreadcrumb[]> | null = null;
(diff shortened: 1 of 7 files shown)


TheRealAgentK:
Converted to draft PR until it's clear where this going in the future.

miquelbeltran:
Incorporated changes into my work branch `miquelbeltran/breadcrumbs`. I'll close this PR and open a new one when ready.

PR #97, opened 2021-01-21
Fix raygunClient.user not being called if request is missing

Widdershin:
This is a fix for a regression I introduced during the TypeScript
overhaul.

Previously we would always call this.user, even if the request was null
or undefined. The regressed behavior would only call this.user if the
request was present, either from the middleware or if it was passed
manually.

This commit reverts the behavior back to what it was previously, and
adds a unit test to guard against future unintended changes to this
behaviour.

### DIFF
--- a/lib/raygun.ts
+++ b/lib/raygun.ts
@@ -122,7 +122,7 @@ class Raygun {
     return this;
   }
 
-  user(req: Request): RawUserData | null {
+  user(req?: Request): RawUserData | null {
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


samuel-holt (approved):
:+1: Looks good, thanks for fixing this!

BenjaminHarding (approved):
👍 Nice fix

Widdershin:
Thanks for the quick review, will get this out as a patch release start of next week.

PR #115, opened 2021-05-26
Fix exception formatting when passing string to send

### LINKED ISSUE #114: Using `raygunClient.send` with a string value results in error messages in Raygun
When passing a string to the send method it results in the message being broken up into comma-separated characters that are prepended with the index of each character.

I have found that setting the `useHumanStringForObject` option to false is a way to get around this issue.

image

Widdershin:
Previously, when calling raygunClient.send("test"), the exception would be formatted as if we had sent raygunClient.send(["t", "e", "s", "t"]).

This was as a result of passing the error value to humanString, which is intending for serializing objects and arrays in a pleasant way.

This change avoids applying any formatting to exceptions passed as a string.

I've also included a release commit so that we can promptly publish after merging.

Fixes #114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.13.1
+- Fix format of exceptions when calling `raygunClient.send` with a string value
+
 ## 0.13.0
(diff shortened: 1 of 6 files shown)


samuel-holt (approved):
:+1: Nice fix, thanks for picking this one up so quickly!

PR #124, opened 2022-03-17
Fix massive errors causing batch processing stalls

Widdershin:
Previously, if we attempted to send a massive error (>1600kb) via batch
mode, it would stall all batch processing.

This can cause all sorts of problems. Now, we dicard any messages this
large and log a warning.

This is better than the alternative behaviour of attempting to send
anyway, as the Raygun API will disconnect us if we intend to send that
much data, which causes an annoying EPIPE error.


grugcrood82 (approved):
LGTM 🚀

Widdershin:
Thanks :)

PR #125, opened 2022-03-22
Send error batches when batch is full

Widdershin:
Previously, we would only send a batch every 1000ms at max, up to one
hundred messages at a time.

This approach has an issue, where if an application generates more
errors more quickly than they're processed, excess errors will accrue in
the messageQueue, wasting memory.

Now, each message is added to the batch as it arrives. Once the batch
hits either the byte limit or message limit, it will be sent.

In addition, when a message arrives, a timer is started to send the
batch after a certain duration. This is to ensure that errors are still
sent promptly even if we don't generate enough to send a full batch.

I've removed the raygun_batch_test suite as the function it tested is
now considered internal, and ultimately this behavior is better tested
in an integration style (see raygun_express_test and
raygun_batch_lifecycle_test).


grugcrood82 (dismissed):
LGTM 🚀

Widdershin:
Hi @grugcrood82, I noticed a few more issues while testing and going over the code prior to merging, would you be able to check these new commits when you have time please?

grugcrood82 (approved):

Widdershin:
Thanks @grugcrood82!

PR #143, opened 2024-04-23
ci: add Dependabot config file

miquelbeltran:
Adds template for Dependabot config, we can change it in the future.


TheRealAgentK (approved):
No brainer, LGTM.

PR #154, opened 2024-04-24
chore: adding PR, bug report and feature request templates

TheRealAgentK:
Adding some process templates as a starting point. We can tweak further down the track, but this will help streamline flows as well as internal/external requests.

### DIFF
--- a/.github/bug_report.md
+++ b/.github/bug_report.md
@@ -30,5 +30,8 @@ If applicable, add screenshots to help explain your problem.
  - SDK version: [e.g. 10.5.2]
  - Node/npm versions: [e.g. 21.7.3]
 
-**Additional context**
-Add any other context about the problem here.
+## Logs
+If applicable, include relevant logs or error messages here (please use code blocks to format properly).
+
+## Possible Solution
+If you have suggestions for how the issue might be resolved or investigated, please describe them here.
--- a/.github/bug_report.md
+++ b/.github/bug_report.md
@@ -30,8 +30,8 @@ If applicable, add screenshots to help explain your problem.
  - SDK version: [e.g. 10.5.2]
  - Node/npm versions: [e.g. 21.7.3]
 
-## Logs
+** Logs**
 If applicable, include relevant logs or error messages here (please use code blocks to format properly).
 
-## Possible Solution
+**Possible Solution**
 If you have suggestions for how the issue might be resolved or investigated, please describe them here.
--- a/.github/bug_report.md
+++ b/.github/bug_report.md
@@ -1,0 +1,34 @@
+---
+name: Bug report
+about: Bug report
+title: "<Area of the SDK>: <Short Description of issue>"
+labels: needs triage
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+Links to related tickets or other context are welcome here.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Environment (please complete the following information):**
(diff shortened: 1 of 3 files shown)


miquelbeltran (approved):

pull_request_template.md
```diff
@@ -0,0 +1,41 @@
+## [{ticket number/type}:] {title}
+
+### Description :memo:
+{Description}
```
sumitramanga, line 4:
```suggestion
- **Purpose**: What issue does this PR address or what feature does it implement?
- **Approach**: How does this PR tackle the problem?
```

pull_request_template.md
sumitramanga, line 1: Is there any links we can add for additional documentation such as our standards/best practises  we follow when requesting a PR that is from the community?

pull_request_template.md
```diff
@@ -0,0 +1,41 @@
...
+
+**Type of change**
+
+*Delete options that are not relevant.*
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
```
sumitramanga, line 12: I never really understood the 'Breaking change ' option 😅 why are we introducing code that would break the provider 🤔 Or am I reading this incorrectly

pull_request_template.md
TheRealAgentK: I'd say yes, but I think we'll need to come up with some structure around that for ourselves first. 

I mean - common expectations would be - add test, tests pass, run linting etc - but let's establish a pattern for ourselves first as part of this work and then document the expectations.

I think these templates are a starting point and not the final outcome.

pull_request_template.md
TheRealAgentK: The way I use "Breaking change" is in the sense of an API contract.

Say for instance - something in the public layer of functions a user would use changes and it breaks backwards compatibility.

Made up example:

Version 1 of SDK: .init(string apiKey) requires the API key 
Version 2 of SDK: .init(string apiKey) is removed and there's no only a .init() and the API is configured elsewhere.

If someone using V1 would just update the dependency to RG4N to V2 in their own application, this change would be breaking their application because we've "broken" the API contract.

pull_request_template.md
sumitramanga: Sounds like a good plan!

.github/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Environment (please complete the following information):**
+ - OS: [e.g. Linux/Windows/macOS etc]
+ - SDK version: [e.g. 10.5.2]
+ - Node/npm versions: [e.g. 21.7.3]
+
+**Additional context**
+Add any other context about the problem here.
```
sumitramanga, line 34:
```suggestion
## Logs
If applicable, include relevant logs or error messages here (please use code blocks to format properly).

## Possible Solution
If you have suggestions for how the issue might be resolved or investigated, please describe them here.
```

pull_request_template.md
sumitramanga: Ahhh, that makes a lot of sense now. Thank you!

.github/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+
+**Environment (please complete the following information):**
+ - OS: [e.g. Linux/Windows/macOS etc]
+ - SDK version: [e.g. 10.5.2]
+ - Node/npm versions: [e.g. 21.7.3]
+
+## Logs
+If applicable, include relevant logs or error messages here (please use code blocks to format properly).
+
+## Possible Solution
```
sumitramanga, line 36:
```suggestion
** Logs**
If applicable, include relevant logs or error messages here (please use code blocks to format properly).

**Possible Solution**
```

Sorry, I got the formatting wrong!

sumitramanga (approved):
LGTM :shipit:

TheRealAgentK:
(sorry folks, did a normal merge instead of a Squash merge because I'm so used to just hitting the button and squash being the default :-) )

PR #156, opened 2024-04-24
ci: #145 eslint setup

miquelbeltran:
### Description :memo:

Removes deprecated `tslint` and setups `eslint`.

- Closes #145 

**Type of change**

- [x] New feature (non-breaking change which adds functionality)

**Updates**

- Removed `tslint` and config file.
- Install `eslint` with support for typescript.
- Setup `eslint` default config.
  - Added rules to support the current codestyle without changes.
- Setup `eslint` command
  - `npm run eslint` to run on `lib` and `test`.
  - Also the files in the `examples` folder are included.
- Setup GitHub Actions to run `eslint`.

Note: The current `eslint` config is set up in a way that no code changes are required. We can fix the code changes and remove the ignored rules in future PRs.

### Test plan :test_tube:

- No code changes done, only dev setup changes.
- CI jobs should pass, which now include running `eslint` step.

### Author to check :eyeglasses:
- [x] Project and all contained modules builds successfully
- [x] Self-/dev-tested 
- [x] Unit/UI/Automation/Integration tests provided where applicable
- [x] Code is written to standards
- [x] Appropriate documentation written (code comments, internal docs)

### Reviewer to check :heavy_check_mark:
- [x] Project and all contained modules builds successfully
- [ ] Change has been dev-/reviewer-tested, where possible
- [ ] Unit/UI/Automation/Integration tests provided where applicable
- [x] Code is written to standards
- [x] Appropriate documentation written (code comments, internal docs)

Issue #148, opened 2024-04-24
Upgrade dependencies on examples


miquelbeltran:
As continuation of the work in https://github.com/MindscapeHQ/raygun4node/pull/144

**TODO:**

Upgrade dependencies on the example projects
- [x] "Express" example
- [x] "Using Domains" example

miquelbeltran:
Completed in #157

PR #157, opened 2024-04-24
chore: #148 upgrade examples

miquelbeltran:
### Description :memo:

Take the two examples in the repo and upgrade dependencies. Also ensure that they use the `raygun4node` found in the repo, not the one from `npm`.

**Type of change**

- [x] Tech debt ;)

**Updates**

- Upgrade dependencies in `examples/express-sample`.
- Upgrade dependencies in `examples/using-domains`.
- Ensure both examples use the `raygun` node package from the repo, not the one hosted in `npm`.
- Adds multiple `console.log` to `examples/using-domains`.
- Fixes indentation in `examples/using-domains/app.js`.

Note: The `eslint` config is done in #156 and includes the files in the `examples` folders.

### Test plan :test_tube:

- Run both examples.

- [x] Self-/dev-tested 

- [x] Change has been dev-/reviewer-tested, where possible

### DIFF
--- a/examples/express-sample/package.json
+++ b/examples/express-sample/package.json
@@ -6,15 +6,15 @@
     "start": "DEBUG=express-sample:server node ./bin/www"
   },
   "dependencies": {
-    "body-parser": "^1.19.0",
-    "config": "^1.16.0",
-    "cookie-parser": "~1.3.3",
-    "debug": "^4.1.1",
-    "ejs": "~2.5.5",
-    "express": "^4.17.1",
+    "body-parser": "^1.20.2",
+    "config": "^3.3.11",
+    "cookie-parser": "~1.4.6",
+    "debug": "^4.3.4",
+    "ejs": "~3.1.10",
+    "express": "^4.19.2",
     "morgan": "^1.10.0",
-    "node-sass-middleware": "^0.11.0",
-    "raygun": "^0.8.0",
+    "node-sass-middleware": "^1.1.0",
+    "raygun": "file:../..",
     "serve-favicon": "^2.5.0"
   }
 }
--- a/examples/express-sample/app.js
+++ b/examples/express-sample/app.js
@@ -1,5 +1,10 @@
 var config = require('config');
 
+if (config.Raygun.Key === 'YOUR_API_KEY') {
+    console.error('You need to set your Raygun API key in the config file');
+    process.exit(1);
+}
+
 // Setup Raygun
 var raygun = require('raygun');
 var raygunClient = new raygun.Client().init({
--- a/examples/express-sample/package-lock.json
+++ b/examples/express-sample/package-lock.json
@@ -1,1364 +1,2762 @@
 {
   "name": "express-sample",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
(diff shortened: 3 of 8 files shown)


examples/express-sample/package.json
```diff
@@ -6,15 +6,15 @@
...
+    "config": "^3.3.11",
+    "cookie-parser": "~1.4.6",
+    "debug": "^4.3.4",
+    "ejs": "~3.1.10",
+    "express": "^4.19.2",
     "morgan": "^1.10.0",
-    "node-sass-middleware": "^0.11.0",
-    "raygun": "^0.8.0",
+    "node-sass-middleware": "^1.1.0",
+    "raygun": "file:../..",
```
miquelbeltran, line 17: `raygun` dependency points to the one at the repo root, instead of using an outdated one from `npm`.

TheRealAgentK (commented):
LGTM overall, haven't built/tested it myself yet though. Will do later and then approve/merge the PR.

sumitramanga (commented):
Code also looks good to me! I'll look to spin this up for testing and follow the instructions

examples/express-sample/package.json
sumitramanga, line 1: It would be awesome to have the README more filled in on how to run the example apps!

I didn't realise we had to build the provider first before building and running the app! This tripped me up badly as I was wondering why TS was needed in the example app as the example app didn't have TS but the provider itself does. Went down a rabbit hold 😬 But realised the situation as mentioned before.

sumitramanga: I think we just need to add in "ensure you've built the provider in the root directory before running this example app". Same goes for the other example app

sumitramanga: Do we need to explicitly point out the API Key change to ensure that the config value is changed to their API Key? Just to make it obvious to the user

sumitramanga (changes requested):
Just one small change that people would benefit having IMO :) 

Otherwise, testing went well and the apps are working with error being piped into Raygun 🙌

sumitramanga (commented):
Oh, one other issue I did have with the express example app was that I had to run `set DEBUG=express-sample:server && node ./bin/www` instead of  `npm start` to get the server up and running. Not sure why, but the server just terminates when running `npm start` so the other option that worked was the manual run in terminal. 

Let me know how you guys get on with `npm start` @miquelbeltran @TheRealAgentK ? It could just be a local issue on my end perhaps...

miquelbeltran:
> I didn't realise we had to build the provider first before building and running the app!

I've added that step now to the README

> Do we need to explicitly point out the API Key change to ensure that the config value is changed to their API Key? Just to make it obvious to the user

I can add that check on the examples, or even in the package itself, that if the key == `YOUR_API_KEY` then it should warn the user.

> Not sure why, but the server just terminates when running npm start so the other option that worked was the manual run in terminal.

I did not have this issue, that's strange...

sumitramanga:
Thank you for addressing those @miquelbeltran !
As for the local set up issues, I don't think this is tied to provider issues themselves but a problem on my computer. So no need to add additional information around how I addressed the problem

sumitramanga (approved):
LGTM! 🎸

PR #169, opened 2024-04-26
ci: #174 Prettier setup

miquelbeltran:
### Description :memo:
- **Purpose**: Enforces code formatting using the `prettier` command.
- **Approach**: Runs the `prettier` command in dry mode in pull-requests.

Depends on #170 

**Type of change**

**Updates**

- Add `prettier` job to the `node.js.yml` GitHub Actions config file.
- Add examples to the `prettier` command in `packages.json`
- Run the `npm run prettier` command (modifies multiple files)

### Test plan :test_tube:

- Prettier reformats code, replaces single quotes by double quotes, etc. Should not change functionality!

- [x] Self-/dev-tested 


miquelbeltran:
image

This is what we will see when the code is not correctly formatted

package.json
```diff
@@ -32,7 +32,7 @@
   ],
   "scripts": {
     "eslint": "eslint --fix lib/*.ts test/*.js examples/**/*.js",
-    "prettier": "prettier --write lib/*.ts test/*.js",
+    "prettier": "prettier --write lib/*.ts test/*.js examples/**/*.js",
     "prepare": "tsc",
     "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js"
```
sumitramanga, line 37: Do we need to add code comments for what these scripts are doing? So that it's easier to understand/digest

.github/workflows/node.js.yml
```diff
@@ -29,3 +29,23 @@ jobs:
     - run: npm ci
     - run: npm test
     - run: npm run eslint
+
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
sumitramanga, line 38: Is it possible to define `@latest` ?

.github/workflows/node.js.yml
```diff
@@ -29,3 +29,23 @@ jobs:
...
+        with:
+          fetch-depth: 0
```
sumitramanga, line 40: would it be more useful to do (this is supposedly faster but I don't know what repercussions it could have) - 
```suggestion
          fetch-depth: 1
```

sumitramanga (commented):
Just left some comments I'd like for clarity before I approve  

I appreciate that this approach adheres to a consistent style without manually formatting each file . Which will save us a lot of time and conversations on how to format what! Especially with this provider being public, the community will know how to format code if they request any PR's

package.json
miquelbeltran: I can do that!

.github/workflows/node.js.yml
miquelbeltran: From the documentation I checked, it doesn't seem to be possible.
https://github.com/orgs/community/discussions/25772

But we have dependabot which will detect updates and apply them.

.github/workflows/node.js.yml
miquelbeltran: Should be 1 indeed!

miquelbeltran: Actually I am wrong, 0 is necessary. Will update the code.

package.json
```diff
@@ -31,9 +31,16 @@
...
-    "prettier": "prettier --write lib/*.ts test/*.js",
+
+    "//prettier": "performs code formatting over project files and examples",
+    "prettier": "prettier --write lib/*.ts test/*.js examples/**/*.js",
+
+    "//prepare": "prepare project for distribution",
     "prepare": "tsc",
+
+    "//test": "runs tests on main project",
     "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js"
```
miquelbeltran, line 44: Adding comments on json files is a bit weird... so I am following this pattern here: https://stackoverflow.com/a/50878067 seems like a good compromise.

.github/workflows/node.js.yml
sumitramanga: I'm happy with this alternative approach!

sumitramanga:
Could you please clarify why we can't do ` fetch-depth: 1` please

package.json
```diff
@@ -31,9 +31,16 @@
...
     "eslint": "eslint --fix lib/*.ts test/*.js examples/**/*.js",
-    "prettier": "prettier --write lib/*.ts test/*.js",
+
+    "//prettier": "performs code formatting over project files and examples",
+    "prettier": "prettier --write lib/*.ts test/*.js examples/**/*.js",
+
+    "//prepare": "prepare project for distribution",
     "prepare": "tsc",
+
+    "//test": "runs tests on main project",
```
sumitramanga, line 43: These comments are much better, thank you!!

sumitramanga (approved):
LGTM 🧶

miquelbeltran:
> Could you please clarify why we can't do  fetch-depth: 1 please

Although the prettier GitHub Action asked for it, I ended up removing it and it still seems to work.

I think it is only necessary if we wanted to apply prettier only to the changes done, and not the whole codebase, so it doesn't seem to be necessary in our case.

PR #160, opened 2024-04-25
refactor: #158 remove no-this-alias ignore eslint rule

miquelbeltran:
### Description :memo:

- Removes the `no-this-alias` rule from the eslint configuration, applies code changes.
- Part of #158 
- Depends on #145 
- Depends on #159

More context on this lint rule: https://typescript-eslint.io/rules/no-this-alias/

> Assigning a variable to this instead of properly using arrow lambdas may be a symptom of pre-ES6 practices or not managing scope well.

**Type of change**

- [x] tech debt

**Updates**

- Removed ignored rule from config file.
- Use of arrow lambdas to keep the constantly bound `this` reference (more info: https://stackoverflow.com/a/56206672/673294)
- In the case of `HttpOptions`, the object is created before returning the `MessageTransport` object.

### Test plan :test_tube:

- Run the `npm ci`, `npm test` and `npm run eslint` commands to ensure no issues raised.
- Verified that there were test cases using those changed functions.
- Run examples.

### Author to check :eyeglasses:
- [x] Project and all contained modules builds successfully
- [x] Self-/dev-tested 
- [x] Unit/UI/Automation/Integration tests provided where applicable
- [x] Code is written to standards
- [x] Appropriate documentation written (code comments, internal docs)

### Reviewer to check :heavy_check_mark:
- [ ] Project and all contained modules builds successfully
- [ ] Change has been dev-/reviewer-tested, where possible
- [x] Unit/UI/Automation/Integration tests provided where applicable
- [x] Code is written to standards
- [x] Appropriate documentation written (code comments, internal docs)

Issue #158, opened 2024-04-24
eslint: remove ignored rules and fix issues


miquelbeltran:
**Please describe your new feature request**

In the PR https://github.com/MindscapeHQ/raygun4node/pull/156 we setup `eslint` to be the linter platform for the project.

However, the configuration file contains several rules set to "off" in order to pass the lint checks without changing the project.

We should eventually enable them and cleanup the code as well.

**Describe the solution you'd like**

Remove the rules' definition from `eslint.config.mjs`:
- [x] `@typescript-eslint/no-explicit-any`
- [x] `@typescript-eslint/no-this-alias`
- [x] `@typescript-eslint/no-unused-vars`
- [x] `@typescript-eslint/no-var-requires`
- [x] `no-undef`
- [ ] `no-unreachable`
- [ ] `no-useless-catch`

In case there are rules that should be there, document why in the `eslint.config.mjs` file.

**Describe alternatives you've considered**

We should not leave those rules there if we plan to continue development with the plugin, as they prevent us from detecting possible issues earlier on.

**Additional context**

There are around ~120 warnings when removing the rules.

miquelbeltran:
Closed by mistake. Still two PRs to go.

PR #170, opened 2024-04-26
refactor: #158 fix eslint rule "no-unused-vars"

miquelbeltran:
### Description :memo:
- **Purpose**: Fixes the "no-unused-vars" lint rule.
- **Approach**: Configured the rule and performed code changes.

To allow unused parameters, this rule sets the property `args` to `none`, that way the linter will not report errors for this case, e.g. here the `req` parameter would throw an error without this configuration:

```js
raygunClient.user = function(req) {
    return 'user@example.com';
};
```

- Depends on #160 

**Type of change**

**Updates**

- Continuing work from pr #160 
- Configure rule as explained.
- Removed several unused variables.

### Test plan :test_tube:

- run ci scripts (test and lint)
- test example apps

- [ ] Self-/dev-tested 


eslint.config.mjs
```diff
@@ -8,8 +8,15 @@ export default tseslint.config(
     ...tseslint.configs.recommended,
     {
         rules: {
+            // Unused vars reported as error
+            "@typescript-eslint/no-unused-vars": [
+                "error",
+                {
+                    // Do not report unused function arguments
+                    "args": "none",
```
sumitramanga, line 16: I was wondering why we wouldn't enforce the rule to error out when unused arguments are declared. We can force the contributor to remove tech debt by having this rule on and minimize the time for this when reviewing PRs.  I would be keen to hear your thoughts :)

sumitramanga (changes requested):
The code changes itself look good! Just one thing I'd like clarified please

eslint.config.mjs
miquelbeltran: The problem I have found is that the error will raise as well in a situation like these:

```js
 expressCustomData(error: Error, request: Request) {
    return {};
  }
```

Here, we define a function that the user can override, but the default implementation doesn't use its parameters, and so `eslint` would complain about that.

In the README, you can see an example use of this, where the two parameters are defined:

```js
raygunClient.expressCustomData = function (err, req) {
  return { 'level': err.level };
};
```

Alternatively, we can add this rule and then add the "ignore" comment on each one of the individual cases where it should be.

sumitramanga: I see now. So this is in regard to the function needed to match the signature specified so we can't just use `expressCustomData ` without the parameters. To my understanding.

I'm happy with this justification :) We will just need to keep an eye on code reviews to ensure we aren't leaving being unused parameters that aren't needed

sumitramanga (approved):
LGTM 🍾

eslint.config.mjs
miquelbeltran: Yes, if we see that this becomes a problem, I'll look for an alternative.

PR #173, opened 2024-04-29
chore: #158 document why no-var-requires ignore is necessary

miquelbeltran:
### Description :memo:
- **Purpose**: Document why the `eslint` rule `no-var-requires` for the correct function of the project.
- **Approach**: Add comment on rule.

After removing the rule ignore, I started converting some imports.

For example, before it would be:

```js
const humanString = require("object-to-human-string");
```

After, it would be:

```js
import humanString from "object-to-human-string";
```

But this leads to errors:

```
lib/raygun.messageBuilder.ts:29:25 - error TS7016: Could not find a declaration file for module 'object-to-human-string'. '/home/miquel/dev/projects/raygun/raygun4node/node_modules/object-to-human-string/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/object-to-human-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'object-to-human-string';`

29 import humanString from "object-to-human-string";
                           ~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in lib/raygun.messageBuilder.ts:29
```

I understand that this is because we cannot import JS modules this way, so decided to keep this rule as it is for now.
I have added a comment explaining why the rule ignore is necessary, and we can move to the next one.

**Type of change**

- [x] Tech-debt

**Updates**

- Updated `eslint.config.mjs` with commentary.

### Test plan :test_tube:

- No actual changes happening

- [x] Self-/dev-tested 


sumitramanga:
Just to confirm - We just stick to doing  `const humanString = require("object-to-human-string");` ?

sumitramanga (approved):
LGTM! I understand now that we use `require` 💫

miquelbeltran:
Correct! we leave it as it is

miquelbeltran:
waiting for merge as it depends on the other PRs

PR #176, opened 2024-04-30
refactor: #158 configure no-undef eslint rule 

miquelbeltran:
### Description :memo:
- **Purpose**: Configure `no-undef` rule to detect use of undefined variables/functions/etc.
- **Approach**: Remove the `off` setting and configure the globals.
- **Depends**: Depends on #173 to be merged

**Type of change**

**Updates**

- Remove `no-undef` `off` setting.
- Configure globals to accept definitions like `console` or `require`.
- Ingore `fakeErrorHandler.DealWith(err);` on purpose.

### Test plan :test_tube:

- Should pass `npm run eslint`.
- No other code changes.

- [x] Self-/dev-tested 


eslint.config.mjs
```diff
@@ -7,6 +7,17 @@ export default tseslint.config(
...
+            // Add node globals to ignore undefined
+            globals: {
+                "__dirname": false,
+                "console": false,
+                "module": false,
+                "process": false,
+                "require": false,
+                "setTimeout": false,
+            }
+        },
```
miquelbeltran, line 20: An alternative to this would be:
- Install the `globals` npm package: https://www.npmjs.com/package/globals
- Import those predefined globals into the config file, e.g.

```mjs
globals: {
  ...globals.node,
}
```

I like defining them as we go, so we have one less dependency, but if this becomes unmanageable, then we can switch to predefined globals.

examples/using-domains/app.js
```diff
@@ -46,6 +46,7 @@ appDomain.run(function () {
       // Or we can deal with it in our "Fake Error Handler" below
 
       // This will throw an error as fakeErrorHandler doesn't exist
+      // eslint-disable-next-line
       fakeErrorHandler.DealWith(err);
```
miquelbeltran, line 50: Ignored on purpose, so this raises a runtime error that will be reported to raygun.

eslint.config.mjs
sumitramanga: I agree on having one less dependency unless the management side outweighs the benefit of less dependencies. Let's leave it as is until it really becomes a problem :)

sumitramanga (approved):
LGTM! 🍎

Issue #146, opened 2024-04-24
Improve async behaviour when sending


TheRealAgentK:
>Ideally I'm after something as simple as `await client.send(err);` (assuming a client is initialised)
>No convoluted promise chains etc.

_Originally posted by @paul-uz in https://github.com/MindscapeHQ/raygun4node/issues/139#issuecomment-2071769894_
            
Note from Kai: see full thread in #139 for more context

TheRealAgentK:
There's additional info and an already existing ticket from the OP in #141.

miquelbeltran:
Resolved in 0.15.0-0

Issue #141, opened 2024-04-11
`.send()` does not with without async/await.


paul-uz:
Documentation leads you to believe that the `.send()` method can simply be called and data will be sent to your dashboard.

Using 0.13.2, this does not seem to be the case. I only get errors in my dashboard if I await the `.send()` method.

Are the docs out of date? Am I doing something wrong?

Here is my basic code, ran inside a Vitest test

```
    const foo = new Raygun.Client().init(
      {
        apiKey: '<MY_API_KEY>',
        reportColumnNumbers: true,
        reportUncaughtExceptions: true,
        batch: false,
        tags: ['vitest'],
      },
    );

    foo.send('test', {}, (res) => {
      console.log(res);
    });
```

miquelbeltran:
Resolved in 0.15.0-0

PR #181, opened 2024-05-02
feat: #146 async send support

miquelbeltran:
### Description :memo:
- **Purpose**: Adds support to async/await to `send` API method
- **Approach**: Preserve the original `send` as `sendWithCallback` and create a new `send` async method.
- Part of: #141 
- Part of: #146 

**Note:** the client requires a major clean-up to fully remove all callbacks from the internal implementation. This will be done in separated PRs after this one.

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Updates**

  - The original `send` with a callback was renamed to `sendWithCallback`, it is still being used internally, so it was hard to fully remove the existing callback parameter without doing a major package rewrite.
  - Created a new `send` method that supports awaited calls.
    - New method returns a `Promise` that can be an `IncomingMessage`. Users can ignore this response if they want, or used to verify that the error was sent correctly to Raygun.
    - If the method fails, the `Promise` will be "rejected", users can catch this error. The error payload can be either `Error & IncomingMessage` or just `Error`.
    - To implement this async method, I used the `Promise` constructor, wrapping the legacy `sendWithCallback` method.

Other changes:

- Reviewed the internal `send` implementations, they still use callbacks internally, and will require a lot of work to clean up. This work should be done in small batches split in several PRs.
- Updated `README.md` with usage examples.
- Updated example `examples/using-domains`.
  - note that the express example doesn't use the `send()` method directly.

### Test plan :test_tube:

- Added a new unit test suite that copies the original `raygun_send_test` but for async `send` calls.
- Updated and tested on the `examples/using-domains`, which now uses the async `send` method.

- [x] Self-/dev-tested 


lib/raygun.ts
```diff
@@ -285,9 +318,11 @@ class Raygun {
       body: req.body,
     };
 
-    this.send(err, customData || {}, function () {}, requestParams, [
+    this.send(err, customData || {}, requestParams, [
       "UnhandledException",
-    ]);
+    ]).catch((err) => {
+      console.log(`[Raygun] Failed to send Express error: ${err}`);
+    });
```
miquelbeltran, line 325: This is internal to the `express` error handler.

test/raygun_offline_test.js
```diff
@@ -17,18 +17,7 @@ test("offline message storage and sending", async function (t) {
...
-      raygunClient.send(e, null, (err, data) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(data);
-        }
-      });
-    });
-
-  await send(new Error("offline error"));
```
miquelbeltran, line 31: this is no longer necessary, because now the `raygunClient.send` method is asynchronous

test/raygun_send_test.js
```diff
@@ -27,7 +27,7 @@ test("send basic", {}, function (t) {
   var client = new Raygun.Client().init({
     apiKey: API_KEY,
   });
-  client.send(new Error(), {}, function (response) {
+  client.sendWithCallback(new Error(), {}, function (response) {
```
miquelbeltran, line 30: The old `raygun_send_test` still uses the legacy `sendWithCallback` method, this test fine can be removed when the `sendWithCallback` method is finally removed.

lib/raygun.sync.transport.ts
```diff
@@ -21,10 +21,17 @@ function syncRequest(httpOptions: SendOptionsWithoutCB) {
...
+ * Spawns a synchronous send request.
+ * Errors are not returned and callback is ignored.
+ * Only used to report uncaught exceptions.
+ * @param options
+ */
 export function send(options: SendOptionsWithoutCB) {
   try {
     syncRequest(options);
   } catch (e) {
+    // TODO: Is there a reason we ignore errors here?
```
sumitramanga, line 34: Perhaps we turn this into an issue?

lib/raygun.transport.ts
```diff
@@ -26,6 +26,12 @@ export function sendBatch(options: SendOptions) {
   return send(options, BATCH_ENDPOINT);
 }
 
+// TODO: Convert this method callbacks to Promise.
```
sumitramanga, line 29: Same as above - Turn into an Issue

lib/raygun.transport.ts
```diff
@@ -66,6 +72,7 @@ export function send(options: SendOptions, path = DEFAULT_ENDPOINT) {
     request.write(data);
     request.end();
   } catch (e) {
+    // TODO: Non-HTTP errors are being ignored, should be better pass them up?
```
sumitramanga, line 75: Same as above - Turn into an Issue

lib/raygun.ts
```diff
@@ -185,10 +186,45 @@ class Raygun {
...
+          }
+        },
+        request,
+        tags,
+      );
+    });
+  }
+
+  /**
+   * @deprecated use send instead, which supports async/await calls.
```
sumitramanga, line 222:
```suggestion
   * @deprecated use sendWithCallback instead, which supports async/await calls.
```

Is this suggestion correct?

sumitramanga (dismissed):
Implementation overall LGTM! Nice work on downsizing the send call 💪

miquelbeltran:
I need a new review... :upside_down_face: 

The tickets have been created, I'll edit them with some extra info.

paul-uz:
What's the ETA on getting this merged and released?

I'm super keen to try it out and get it hooked into Winston as a custom transport.

miquelbeltran:
Hi @paul-uz we are talking internally that we want to do a pre-release with these changes once they are merged in. I'll try to remember to ping you when we have it ready, and be looking forward to your feedback!

paul-uz:
Please do what you can to extradite this ✌️

sumitramanga:
Just double checking - Did you turn all to do's into an Issue?

sumitramanga (approved):
Thank you for the changes and making those Issues! LGTM🍾

paul-uz:
@miquelbeltran how long until a new release version is out?

PR #179, opened 2024-05-01
chore: Prepare for Release 0.14.0

miquelbeltran:
### Description :memo:
- **Purpose**: Prepare package for release
- **Approach**: Bump version number and update changelog.

**Type of change**

- [x] Release

**Updates**

- Sets version to `0.14.0`
- Adds Changelog.

### Author to check :eyeglasses:
- [ ] Project and all contained modules builds successfully
- [ ] Self-/dev-tested 
- [ ] Unit/UI/Automation/Integration tests provided where applicable
- [ ] Code is written to standards
- [ ] Appropriate documentation written (code comments, internal docs)

### Reviewer to check :heavy_check_mark:
- [x] Project and all contained modules builds successfully
- [x] Change has been dev-/reviewer-tested, where possible
- [x] Unit/UI/Automation/Integration tests provided where applicable
- [x] Code is written to standards
- [x] Appropriate documentation written (code comments, internal docs)

PR #182, opened 2024-05-02
chore: Release 0.14.0

miquelbeltran:
### Description :memo:
- **Purpose**: Release 0.14.0
- **Approach**: Merge to `master`

**Note: DO NOT Squash and Merge, but Rebase!**

**Type of change**

- [x] Release

**Updates**

- [ ] Self-/dev-tested 

- [x] Change has been dev-/reviewer-tested, where possible


sumitramanga (dismissed):
I've tested and code reviewed during the previous release branch (https://github.com/MindscapeHQ/raygun4node/pull/179) which piggy backed off of all the changes in `develop` branch that needs to be merged in. So I'll go ahead and approve this PR 

Great work team!! 💪🏽

miquelbeltran:
ugh, first time I see this :)

image

sumitramanga:
@miquelbeltran that happens when a commit is pushed to the branch 😅 But that won't happen when merging master into your branch has no conflicts

sumitramanga (approved):
Testing went well! LGTM 🚀

TheRealAgentK (dismissed):
LGTM

examples/express-sample/app.js
```diff
@@ -1,52 +1,60 @@
-var config = require('config');
+var config = require("config");
+
+if (config.Raygun.Key === "YOUR_API_KEY") {
```
PanosNB, line 3: Is this the only verification we should be doing here? We could also check expected format with pattern-matching for example?

PanosNB: Ok, I see this is for an example app, so worries

lib/raygun.messageBuilder.ts
```diff
@@ -31,23 +31,26 @@ type UserMessageData = RawUserData | string | undefined;
...
+  // Make temporary copy of the object to avoid mutating the original
+  // Cast to Record<string, object> to enforce type check and avoid using any
+  const _obj = { ...obj } as Record<string, object>;
   Object.keys(obj).forEach(function (i) {
     if (filters.indexOf(i) > -1) {
-      delete obj[i];
+      delete _obj[i];
     } else {
-      obj[i] = filterKeys(obj[i], filters);
+      _obj[i] = filterKeys(_obj[i], filters);
```
PanosNB, line 45: What will happen if there a cycle in the object? Will we get a Stack Overflow if an object down the line, points back to a parent object? Can we add a test for this, if none exists?

lib/raygun.offline.ts
```diff
@@ -29,15 +28,14 @@ export class OfflineStorage implements IOfflineStorage {
...
-      cacheContents
-    ) {
-      storage.transport.send(cacheContents);
-      fs.unlink(path.join(storage.cachePath, item), () => {});
-    });
+    fs.readFile(
+      path.join(this.cachePath, item),
+      "utf8",
+      (err, cacheContents) => {
+        this.transport.send(cacheContents);
```
PanosNB, line 35: Do we need to check if the transport has worked before deleting? Maybe, put the delete in a promise, if send succeeds?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "raygun",
-  "description": "Raygun.io package for Node, written in TypeScript",
-  "version": "0.13.2",
+  "description": "Raygun package for Node.js, written in TypeScript",
+  "version": "0.14.0",
   "homepage": "https://github.com/MindscapeHQ/raygun4node",
   "author": {
     "name": "MindscapeHQ",
```
PanosNB, line 7: Please update all "MindscapeHQ" references to "Raygun" :-)

PanosNB (commented):
Great work team! Really enjoyed the great Software Engineering process structure going in there! Just a few points to consider and a pedantic task

package.json
TheRealAgentK: It might be worthwhile checking if there's a connection between `author.name` in package.json and the account that RG is using to publish to npm. I'm not sure if there's a requirement for them to match maybe?

lib/raygun.messageBuilder.ts
TheRealAgentK: That is a good point. I have seen a similar (in the grand scheme of things) problem in the past with another provider where error objects and structures have all of a sudden introduced unexpected self-referencing or issues when serialising.

:+1:  for test

examples/express-sample/app.js
TheRealAgentK: Curious because you brought it up - is there a format for RG API keys? Because if there is, then it'd be useful to add that to other providers as well.

lib/raygun.messageBuilder.ts
miquelbeltran: Is this a blocker for the release?

lib/raygun.offline.ts
miquelbeltran: The package functionality has not been changed in this case. I am currently working on adding async/await support to the `send` call, which would make easier to do this check, otherwise I'd have to use a callback.

Can we wait for the next release to do this?

package.json
miquelbeltran: Author name doesn't appear on npm even https://www.npmjs.com/package/raygun

The documentation doesn't mention any requirement for them to match https://docs.npmjs.com/cli/v6/configuring-npm/package-json#people-fields-author-contributors

It's a quick change nevertheless so I can do that.

examples/express-sample/app.js
miquelbeltran: This verification is added to the example projects to warn the user that they did not change the default one.

Marking this conversation as resolved as doesnt need an action.

lib/raygun.messageBuilder.ts
miquelbeltran: `filterKeys` is only being used to filter the debug information from the http request, as seen here:

```
queryString: filterKeys(request.query, this._filters),
headers: filterKeys(request.headers, this._filters),
form: filterKeys(request.body, this._filters),
```

It is not being used on error structures, but only on the original request query, headers and body, so I don't expect a recursive object there.

Please let me know if this needs to be researched now or should I create a ticket to look it up in the future.

package.json
miquelbeltran: PR: https://github.com/MindscapeHQ/raygun4node/pull/185

lib/raygun.messageBuilder.ts
TheRealAgentK: Solution: we all agreed to not deal with this potential (albeit unlikely) issue and will look into it post-release.

lib/raygun.offline.ts
TheRealAgentK: Resolution: Agreement on looking at this issue again in the context of the async/await changes.

PanosNB (approved):
Thanks @Miguel Beltran! It seems that the recursive object error is unlikely to surface easily. So yes, you can proceed with this release but let's ensure we protect against recursive objects in the next release please. Same for the other one. And thanks for taking care of renaming MindscapeHQ to Raygun!

TheRealAgentK (approved):
Re-approval after additional commit.

lib/raygun.offline.ts
TheRealAgentK: To be looked at again in the context of https://github.com/MindscapeHQ/raygun4node/issues/146

sumitramanga (approved):
LGTM! Example apps are working well

Issue #186, opened 2024-05-03
Look into potential cyclic objects


TheRealAgentK:
This stems from a PR comment from @panosNB:

 
> What will happen if there a cycle in the object? Will we get a Stack Overflow if an object down the line, points back to a parent object? Can we add a test for this, if none exists?

_Originally posted by @PanosNB in https://github.com/MindscapeHQ/raygun4node/pull/182#discussion_r1588529243_

PR #187, opened 2024-05-03
fix: #186 avoid infinite recursion in filter method

miquelbeltran:
### Description :memo:
- **Purpose**: `filterKeys` method uses recursion to explore the provided object, this can potentially cause infinite recursion and eventually throw the "Maximum call stack size exceeded" error.
- **Approach**: Adds an `explored` parameter to the recursive `filterKeys` method, to check if the incoming object has already been explored or not.

**Type of change**

**Updates**

- TDD approach: Created a test that can reproduce the issue, using a cyclic object, and causes a "Maximum call stack size exceeded" error, then implemented the solution to fix it.
- Implementation:
  - Added `explored: Set<object>` as parameter (it is `null` by default).
  - When iterating over the object children, if the child is in the explored `Set`, remove from parent. Same as when applying filters.
  - When calling to `filterKeys` recursively, include the explored object in the `Set`.
- After performing the changes, the test passes.
- Also added a test assert to ensure that the original object is not modified.

One decision I had to make, is if the self-referenced object should be included in the output or not.

My decision to remove it from the output, is that we cannot apply filtering to the properties of the self-referenced object. Remember that we are doing a copy because we don't want to modify the original.

example of input from the test:

```
{
  myself: *self reference*,
  key: "value",
  filtered: true,
  other: {
    body: *self reference*,
    filtered: true,
  }
}
```

output would be:
```
{
  key: "value",
  other: { }
}
```

After removing all filtered properties and self-referenced objects.

### Test plan :test_tube:

- Method is unit tested with a new test, also existing tests pass.

- [x] Self-/dev-tested 

### DIFF
--- a/lib/raygun.messageBuilder.ts
+++ b/lib/raygun.messageBuilder.ts
@@ -31,28 +31,40 @@ type UserMessageData = RawUserData | string | undefined;
 const humanString = require("object-to-human-string");
 const packageDetails = require("../package.json");
 
+/**
+ * Filter properties in obj according to provided filters.
+ * Also removes any recursive self-referencing object.
+ * @param obj object to apply filter
+ * @param filters list of keys to filter
+ * @param explored Set that contains already explored nodes, used internally
+ */
 function filterKeys(
   obj: object,
   filters: string[],
   explored: Set<object> | null = null,
 ): object {
-  // check if obj has been explored to avoid infinite recursion
-  if (!obj || !filters || typeof obj !== "object" || explored?.has(obj)) {
+  if (!obj || !filters || typeof obj !== "object") {
     return obj;
   }
+
+  // create or update the explored set with the incoming object
+  const _explored = explored?.add(obj) || new Set([obj]);
 
   // Make temporary copy of the object to avoid mutating the original
   // Cast to Record<string, object> to enforce type check and avoid using any
   const _obj = { ...obj } as Record<string, object>;
-  Object.keys(obj).forEach(function (i) {
-    if (filters.indexOf(i) > -1) {
-      delete _obj[i];
+
+  Object.keys(obj).forEach(function (key) {
+    // Remove child if:
+    // - the key is in the filter array
+    // - the value is already in the explored Set
+    if (filters.indexOf(key) > -1 || _explored?.has(_obj[key])) {
+      delete _obj[key];
     } else {
-      // add original obj to explored set before recursive call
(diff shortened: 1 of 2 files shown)


lib/raygun.messageBuilder.ts
```diff
@@ -31,18 +31,30 @@ type UserMessageData = RawUserData | string | undefined;
 const humanString = require("object-to-human-string");
 const packageDetails = require("../package.json");
 
-function filterKeys(obj: object, filters: string[]): object {
-  if (!obj || !filters || typeof obj !== "object") {
+function filterKeys(
+  obj: object,
+  filters: string[],
+  explored: Set<object> | null = null,
```
miquelbeltran, line 37: `explored` is set to null by default, so it is not a mandatory parameter

test/raygun.messageBuilder_test.js
```diff
@@ -377,6 +377,10 @@ test("filter keys tests", function (t) {
   });
   var message = builder.build();
 
+  // Original object should not be modified
+  t.equal(body.username, "admin@raygun.io");
+  t.equal(body.password, "nice try");
+
```
miquelbeltran, line 383: Added to ensure that we do the copy in the `filterKeys` method, i.e. `const _obj = { ...obj } as Record<string, object>;`

TheRealAgentK (approved):
This looks good to me!

lib/raygun.messageBuilder.ts
```diff
@@ -31,18 +31,37 @@ type UserMessageData = RawUserData | string | undefined;
...
 
-function filterKeys(obj: object, filters: string[]): object {
+/**
+ * Filter properties in obj according to provided filters.
+ * Also removes any recursive self-referencing object.
+ * @param obj object to apply filter
+ * @param filters list of keys to filter
+ * @param explored Set that contains already explored nodes, used internally
+ */
+function filterKeys(
```
sumitramanga, line 41: Not sure if it's been checked but it may be useful to check around the repo to see if we've done infinite recursion in the past and fix that up?

sumitramanga (approved):
LGTM! Just one comment to address this issue as a whole in the repo if that hasn't been done yet :)

lib/raygun.messageBuilder.ts
miquelbeltran: I couldn't see any other instances in this repo. Do you mean in all the codebases from Raygun?

Issue #205, opened 2024-05-08
codestyle: consistent use of backticks (string templates), quotes and double quotes.


miquelbeltran:
**Please describe your new feature request**

Strings in Javascript can be defined using backticks, quotes and double quotes, e.g.

```js
`backtick`
'quote'
"double quote"
```

My understanding is that backticks are for string interpolation and other string operations, and for normal strings either double or single quotes are preferred.

**Describe the solution you'd like**

eslint doesn't check for this automatically, but rather an extra rule is required. 

I suggest using https://eslint.style/rules/js/quotes

**Describe alternatives you've considered**

- TBD

**Additional context**

AirBnB codestyle guide: https://github.com/airbnb/javascript#strings

miquelbeltran:
closed in #207

PR #207, opened 2024-05-09
refactor: #205 Setup eslint style

miquelbeltran:
### Description :memo:
- **Purpose**: Provided consistency with the use of quotes, semicolons, etc.
- **Approach**: Setup eslint style as a complement to eslint

**Type of change**

- [x] Refactor

**Updates**

- Installs stylistic eslint and ts plugins.
- Configure eslint.
- apply code changes.

stylistic-eslint and prettier can "disagree" sometimes, that's why some rules have been disabled.

### Test plan :test_tube:

- Code remains functionally the same, GitHub CI should run without complains

- [ ] Self-/dev-tested 

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@eslint/js": "^9.2.0",
     "@stylistic/eslint-plugin": "^2.0.0",
+    "@stylistic/eslint-plugin-ts": "^2.0.0",
     "@types/node": "^20.12.8",
     "@types/stack-trace": "0.0.33",
     "@types/uuid": "^9.0.8",
--- a/package.json
+++ b/package.json
@@ -33,18 +33,16 @@
   "scripts": {
     "//eslint": "performs static analysis over project files and examples",
     "eslint": "eslint --fix lib/*.ts test/*.js examples/**/*.js",
-
     "//prettier": "performs code formatting over project files and examples",
     "prettier": "prettier --write lib/*.ts test/*.js examples/**/*.js",
-
     "//prepare": "prepare project for distribution",
     "prepare": "tsc",
-
     "//test": "runs tests on main project",
     "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js"
   },
   "devDependencies": {
     "@eslint/js": "^9.2.0",
+    "@stylistic/eslint-plugin": "^2.0.0",
     "@types/node": "^20.12.8",
     "@types/stack-trace": "0.0.33",
     "@types/uuid": "^9.0.8",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -33,18 +33,17 @@
...
     "//eslint": "performs static analysis over project files and examples",
     "eslint": "eslint --fix lib/*.ts test/*.js examples/**/*.js",
-
     "//prettier": "performs code formatting over project files and examples",
     "prettier": "prettier --write lib/*.ts test/*.js examples/**/*.js",
-
     "//prepare": "prepare project for distribution",
     "prepare": "tsc",
-
     "//test": "runs tests on main project",
```
miquelbeltran, line 40: I think the blank lines were removed by the npm tool when I ran `npm -i -D ....`

TheRealAgentK:
I don't see why quotes would be preferred over backticks apart from the AirBnb style guide saying so, to be honest.

Many yrs ago, backticks used to be slower, but essentially seem to have been offering the same performance for a long time. They're more flexible and support multi-line and kind-of represent modern JS.

TheRealAgentK:
> I don't have a preference, but I'd prefer a consistent style that is enforced by the static analysis tools.

Fully agreeing!

> Using backticks everywhere in the codebase looks like a bit too much to me, but it is doable automatically, we only need to change the configuration of the rule https://eslint.org/docs/latest/rules/quotes to be `["error", "backtick"]`.

^ This is the only point of discussion, really. 

Here's one list of pros/cons: 

https://medium.com/@sathiranipun/unveiling-the-magic-of-backticks-single-quotes-and-double-quotes-in-javascript-9ef33d32b9fe

Someone looking into the performance (in a browser though): https://dev.to/efpage/the-fantastic-speed-of-template-literals-4f0p

There are obviously numerous punters with opinions out there, but I don't see the drawback of using backticks exclusively at this stage. They're more flexible and they are the more modern style of doing things.

Note: this is *not* the hill I'm going to die on, but I'm curious to what others think. :smile: 

> There is another option, which is to configure the rule as this: `["error", "double", { "allowTemplateLiterals": true }]` this allows you to either chose with double quotes or backticks freely. See https://eslint.org/docs/latest/rules/quotes#allowtemplateliterals
> 
> Alternatively, we can completely disable this rule. But I'd prefer to at least have the above to enforce either double quotes or backticks.
> 
> What do you think @TheRealAgentK ?

I don't want to over-complicate things, let's not complicate rules and just make a decision on either consistent approach.

miquelbeltran:
The good thing about eslint is that running the alias `npm run eslint` will apply fixes automatically (as it has the `--fix` param) so will change any wrong doings.

I will play a bit with the rule config to see if setting backticks is doable and how it will look like.

miquelbeltran:
Another thing to keep in mind is that backticks cannot be used for property values, so we will have to define another rule if we want property values as single or double quotes, e.g. 
```js
var obj = { 'prop-name': `value` };
```

TheRealAgentK (approved):
LGTM after an internal discussion on how we want to handle this.

sumitramanga (approved):
For visibility, I'm happy with this too :)

Issue #74, opened 2018-11-30
No breadcrumb support?


StevePotter:
We are using raygun4node inside AWS Lambda functions.  I would love to use the breadcrumbs feature of Raygun.  It's in the Ruby library but I see it nowhere in here.  Am I missing something?  I can't use raygun4js because it can't run inside node.

mitchmeyer1:
2nd! Same problem!

PR #210, opened 2024-05-10
feat: Breadcrumbs

miquelbeltran:
### Description :memo:
- **Purpose**: Add breadcrumbs support
- **Approach**: Took the work on the old branch `breadcrumbs` and ported it to the latest codebase.
- Closes https://github.com/MindscapeHQ/raygun4node/issues/74

**Type of change**

**Updates**

- Started from https://github.com/MindscapeHQ/raygun4node/pull/96

The main idea is that breadcrumbs are automatically scoped to the current request in ExpressJS.
Breadcrumbs are also added automatically with the request information. 
For that, users must add the `expressHandlerBreadcrumbs` middleware to their ExpressJS app:

```js
app.use(raygun.expressHandlerBreadcrumbs);
```

Then, users can also add their own custom breadcrumbs calling to `addBreadcrumb()` in the scope of a request.

```js
app.get("/endpoint", (req, res) => {
  // ... inside a request ...
  raygun.addBreadcrumb("My custom breadcrumb");
  // ...
```

As well, users can continue using the Raygun ExpressJS handler:

```js
app.use(raygunClient.expressHandler);
```

When an error is captured by the `expressHandler`, or the user calls to `raygun.send()`, the breadcrumbs in the current scope will be attached to the message details payload automatically.

At any point in time, users have the option to discard all the stored breadcrumbs if they so desire:

```
raygun.clearBreadcrumbs();
```

Besides supporting scope by request, breadcrumbs are added to a global scope when calling `addBreadcrumb()` outside the scope, and will be included in any error sent outside of request scopes.

Internally, the implementation uses `AsyncLocalStorage` to handle scopes within requests.
Quoting the original PR:

> Currently built on top of AsyncLocalStorage, which means that we can run a section of code with a particular scope for breadcrumbs, which allows us to process multiple requests in parallel and still attribute breadcrumbs correctly without a great deal of user intervention.

#### Expanded Express example

The ExpressJS example has been expanded this way:

- Added breadcrumbs to project setup.
- Created a `/send` endpoint which sends a custom Raygun error.
- Added breadcrumbs to `/error` endpoint.

Example of how Breadcrumbs appear:

image

As raw data:

```json
"breadcrumbs": [
        {
            "category": "http",
            "message": "GET /send",
            "level": "info",
            "timestamp": 1715668231800,
            "type": "request"
        },
        {
            "level": "debug",
            "category": "Example",
            "message": "Breadcrumb in /send endpoint",
            "customData": {
                "custom-data": "data"
            },
            "timestamp": 1715668231804,
            "type": "manual",
            "className": "/home/miquel/dev/projects/raygun/raygun4node/examples/express-sample/routes/index.js",
            "methodName": "<anonymous>",
            "lineNumber": 14
        }
    ],
```

### Test plan :test_tube:

- Added tests in `test/raygun_breadcrumbs_test.js`.
- Added breadcrumbs to ExpressJS example project.
- Extended the ExpressJS example project.
- Added breadcrumbs to domains test.

- [x] Self-/dev-tested 


miquelbeltran:
Hi @MindscapeHQ/kai-miquel-and-friends I'd like you to take a look at the current implementation while I am finishing documentation and examples, and get some feedback and questions from you.

lib/breadcrumbs.ts
```diff
@@ -0,0 +1,182 @@
...
+    category: breadcrumb.category || "",
+    message: breadcrumb.message || "",
+    level: breadcrumb.level || "info",
+    timestamp: Number(new Date()),
+    type,
+    className: callsite?.fileName,
+    methodName: callsite?.functionName,
+
+    // TODO - do we need to do any source mapping?
+    lineNumber: callsite?.lineNumber || undefined,
```
miquelbeltran, line 96: Comment was in the original implementation of `breadcrumbs.ts`

lib/breadcrumbs.ts
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+  debug("[breadcrumbs.ts] recorded breadcrumb:", internalCrumb);
+
+  crumbs.push(internalCrumb);
+}
+
+export function addRequestBreadcrumb(request: Request, response: Response) {
```
miquelbeltran, line 104: logs an ExpressJS request as a Breadcrumb, used by the `expressHandlerBreadcrumbs`

lib/breadcrumbs.ts
```diff
@@ -0,0 +1,182 @@
...
+  store: InternalBreadcrumb[] = [],
+) {
+  if (!asyncLocalStorage) {
+    f();
+    return;
+  }
+  debug("running function with breadcrumbs");
+
+  asyncLocalStorage.run(store, f);
+}
```
miquelbeltran, line 154: Scopes the `AsyncLocalStore` to the provided function `f`, used by the `expressHandlerBreadcrumbs`

lib/breadcrumbs.ts
```diff
@@ -0,0 +1,182 @@
...
+// for (const [method, level] of consoleMethods) {
+//   const oldMethod = (console as any)[method];
+//   (console as any)[method] = function logWithBreadcrumb<T>(
+//     this: T,
+//     ...args: any[]
+//   ) {
+//     addBreadcrumb({ message: args.join(" "), level }, "console");
+//     return oldMethod.apply(this, args);
+//   };
+// }
```
miquelbeltran, line 182: Old code from the original `breadcrumb.ts`. It overrites the `console.` methods so they log automatically into breadcrumbs. I'll probably end up removing it, users should be able to do that themselves.

lib/raygun.ts
```diff
@@ -335,6 +352,31 @@ class Raygun {
...
+   * Then call to `raygun.addBreadcrumb(...)` to add breadcrumbs to the future Raygun `send` call.
+   * @param req
+   * @param res
+   * @param next
+   */
+  expressHandlerBreadcrumbs(req: Request, res: Response, next: NextFunction) {
+    breadcrumbs.runWithBreadcrumbs(() => {
+      breadcrumbs.addRequestBreadcrumb(req, res);
+      // Make the current breadcrumb store available to the express error handler
+      res.locals.breadcrumbs = breadcrumbs.getBreadcrumbs();
```
miquelbeltran, line 367: This allows to obtain the scoped breadcrumbs store in the `expressHandler` method, as described below.

lib/raygun.ts
```diff
@@ -355,11 +397,30 @@ class Raygun {
...
 
-    this.send(err, customData || {}, requestParams, [
-      "UnhandledException",
-    ]).catch((err) => {
-      console.error("[Raygun] Failed to send Express error", err);
-    });
+    // If a local store of breadcrumbs exist in the response
+    // run in scoped breadcrumbs store
+    if (res.locals && res.locals.breadcrumbs) {
+      breadcrumbs.runWithBreadcrumbs(() => {
```
miquelbeltran, line 403: Obtain the scoped breadcrumbs from the response locals as explained in the previous comment. They are provided to the send method using the `runWithBreadcrumbs` scoping method.

lib/types.ts
```diff
@@ -194,3 +195,19 @@ export function callVariadicCallback<T>(
 }
 
 export type Callback<T> = CallbackNoError<T> | CallbackWithError<T>;
+
+export type InternalBreadcrumb = {
```
TheRealAgentK, line 199: I find the name of the type slightly misleading. `InternalBreadcrumb` to me sounds more that there's somewhat of a provider-internal and a provider-external breadcrumb. I would probably name this more along the lines of `Breadcrumb` (the Partial) and `BreadcrumbInternals` oder `BreadcrumbDetails` to focus more on the fact that both are around the same type in the grand scheme of things.

lib/types.ts
```diff
@@ -194,3 +195,19 @@ export function callVariadicCallback<T>(
...
+  category: string;
+  message: string;
+  customData?: CustomData;
+  className?: string;
+  methodName?: string;
+  lineNumber?: number;
+};
+
+export type Breadcrumb = Partial<
+  Pick<InternalBreadcrumb, "level" | "category" | "message" | "customData">
```
TheRealAgentK, line 212: I haven't seen yet where and how the two types are used, but now realising that `Breadcrumb` is a partial pick type of `IntnernalBreadcrumb` I wonder if the naming should be even flipped around so that the complete type is `Breadcrumb` and the partial-picked one gets a name that expresses that it's a light version of a `Breadcrumb`.

Will come back to this comment once I have a clearer picture, but don't want to lose this thought.

lib/raygun.ts
```diff
@@ -24,8 +25,9 @@ import {
   Tag,
   Transport,
 } from "./types";
+import * as breadcrumbs from "./breadcrumbs";
```
TheRealAgentK, line 28: This is deviating from the more commonly used `raygun.<something>.ts` scheme.

lib/raygun.ts
```diff
@@ -335,6 +352,31 @@ class Raygun {
...
 
+  /**
+   * Attach as express middleware to create a breadcrumb store scope per request.
+   * e.g. `app.use(raygun.expressHandlerBreadcrumbs);`
+   * Then call to `raygun.addBreadcrumb(...)` to add breadcrumbs to the future Raygun `send` call.
+   * @param req
+   * @param res
+   * @param next
+   */
+  expressHandlerBreadcrumbs(req: Request, res: Response, next: NextFunction) {
```
TheRealAgentK, line 363: Just a general question - I obviously get that Express is very popular and commonly used, but what are the criteria for a specific 3rd party framework/tooling to get "special" consideration in a provider SDK?

lib/breadcrumbs.ts
```diff
@@ -0,0 +1,182 @@
+import type { AsyncLocalStorage } from "async_hooks";
+import type { Breadcrumb, InternalBreadcrumb } from "./types";
+import type { Request, Response } from "express";
+const debug = require("debug")("raygun").bind(null, "[breadcrumbs]");
+
+let asyncLocalStorage: AsyncLocalStorage<InternalBreadcrumb[]> | null = null;
```
TheRealAgentK, line 6: I had a quick look into the man page for `AsyncLocalStorage` - but just to double-check: this is in-memory thread-local storage, correct? Like what I'd do with `ThreadLocal` in Java?

lib/breadcrumbs.ts
TheRealAgentK, line 1: Can we make sure that for `debug` we spit out the file name as in this example and capitalise the first char?

```
debug(`[raygun.batch.ts] Batch transport - stopping`);
```

TheRealAgentK: (ignore the backticks, that's copied from my OG PR for it, eslint might have change it by now)

lib/breadcrumbs.ts
```diff
@@ -0,0 +1,182 @@
...
+  asyncLocalStorage = new (require("async_hooks").AsyncLocalStorage)();
+  debug("initialized successfully");
+} catch (e) {
+  debug(
+    "failed to load async_hooks.AsyncLocalStorage - initialization failed\n",
+    e,
+  );
+}
+
+type SourceFile = {
```
TheRealAgentK, line 18: Curious why this is declared here and not in `types.ts` ?

lib/breadcrumbs.ts
TheRealAgentK: Should this express-specific behaviour be in a generic `breadcrumbs` module then or maybe in a separate one for handling Express?

TheRealAgentK: Or should at least the function name be less generic then?

lib/breadcrumbs.ts
TheRealAgentK: I think I know what this flow (from `raygun.ts`) tries to (and probably does) achieve. But I'm not clear about the mechanics around it. Are we essentially attaching ourselves to the thread or to the thread's error handler?

lib/breadcrumbs.ts
TheRealAgentK: I think at first glance this looks like a nice feature that could be offered via a setting/config.

lib/types.ts
TheRealAgentK: Ok, I think I'd essentially I'd rejig the naming of these types.

To me, the full type is a `Breadcrumb` and the picked one is something "different" as indicated in my previous comment.

TheRealAgentK (commented):
I think in general this looks good. 

Mainly I have a bunch of questions to better understand some bits around the expressJS ties and I think we should look at the naming of the types.

examples/express-sample/app.js
```diff
@@ -1,24 +1,10 @@
...
-  apiKey: config.Raygun.Key,
-});
-
 var express = require("express");
 var path = require("path");
 var logger = require("morgan");
 var cookieParser = require("cookie-parser");
 var bodyParser = require("body-parser");
 var sassMiddleware = require("node-sass-middleware");
+var raygunClient = require("./logger");
```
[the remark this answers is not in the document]
miquelbeltran: Probably not, I was thinking to rename the file module to `raygun.js` instead. Just to have an easy way to share the Raygun instance.

lib/types.ts
miquelbeltran: This `Breadcrumb` object is what the end user will see, while `InternalBreadcrumb` is what we send after adding some data. 

Users never see the `InternalBreadcrumb` directly, but I agree that it can be renamed to something else.

I'd keep the `Breadcrumb` exported type name as it is.

lib/raygun.ts
miquelbeltran: Asked internally on Slack and the response I got was to focus on ExpressJS at the moment.

lib/breadcrumbs.ts
miquelbeltran: Yes, not thread based, but rather the context is inside the `AsyncLocalStorage.run` function.
The `store` is indeed in memory, in this case an array of `InternalBreadcrumb`.

lib/breadcrumbs.ts
miquelbeltran: Not really a thread as in the same sense Java threads, but in the chain of callbacks and Promises.

This method is added to the ExpressJS middleware chain, so when a request happens, it goes through it, and all code executed internally is inside this `AsyncLocalStorage` context. So all calls to get/add breadcrumbs use the scoped store.

lib/breadcrumbs.ts
miquelbeltran: Looking a bit into it, I think that `Request` is an object from the standard JS library and not an Express object per se, although this method is only used by the `expressHandlerBreadcrumbs` method, it could be used to register non-ExpressJS requests.

We could probably expose an `addRequestBreadcrumb` as well in the `raygun.js` to be used in other web frameworks, if there is a use for it.

lib/breadcrumbs.ts
miquelbeltran: At the top of the file, the `debug` is specified this way:

```
const debug = require("debug")("raygun").bind(null, "[breadcrumbs]");
```

This adds the `[breadcrumbs]` automatically when calling to `debug(something)`. 

However, I need to change the `[breadcrumbs]` to the actual file name, which should be `raygun.breadcrumbs.js`.

TheRealAgentK:
No, that's different, isn't it.

[breadcrumbs] is the tag for the debug module and that can be quite
generic. Express adds something to it for instance that is not related to
the file at all. I added the filename as an additional information for
folks to easily find the file a debug statement from within the provider
comes from.

lib/breadcrumbs.ts
miquelbeltran: Moved to `raygun.breadcrumbs.express.ts`

lib/raygun.ts
miquelbeltran: renamed to `raygun.breadcrumbs.ts`

miquelbeltran:
> No, that's different, isn't it. [breadcrumbs] is the tag for the debug module and that can be quite generic. Express adds something to it for instance that is not related to the file at all. I added the filename as an additional information for folks to easily find the file a debug statement from within the provider comes from.

The output looks correct to me:

```
2024-05-14T12:21:08.984Z raygun [raygun.ts] Client initialized
2024-05-14T12:21:08.989Z raygun [raygun.breadcrumbs.ts] enter with new store
```

Nevertheless, I will change it to follow the rest of the files.

lib/breadcrumbs.ts
miquelbeltran: Refactored to match the same format as the rest of the project.

examples/express-sample/app.js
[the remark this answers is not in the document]
miquelbeltran: Renamed to `raygun.client.js`, so it doesn't conflict with the `raygun` package import

examples/express-sample/routes/index.js
```diff
@@ -1,14 +1,50 @@
-var express = require("express");
-var router = express.Router();
+const express = require("express");
+const router = express.Router();
+const raygunClient = require("../logger");
```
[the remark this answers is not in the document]
miquelbeltran: Renamed to `raygun.client.js`

examples/express-sample/logger.js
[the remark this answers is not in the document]
miquelbeltran: Also needed a way to properly share the raygun client across the app :)

lib/types.ts
miquelbeltran: Looking at the Raygun4Flutter implementation, the public class is `RaygunBreadcrumbMessage`, so I can use that as well.

miquelbeltran: I went with `BreadcrumbMessage` for the public type, and `Breadcrumb` for the internal payload sent to Raygun.

miquelbeltran:
Did some more cleanup, updated README with Breadcrumb documentation, and applied many of the suggestions from Kai.

I mark this now as ready for review, I think it is in a state which is usable and good enough to go as first implementation.

lib/breadcrumbs.ts
TheRealAgentK: To move things along, maybe create a spin-off ticket to track my "feature suggestion" and then remove the commented code from here for this PR's purpose.

miquelbeltran: Done: https://github.com/MindscapeHQ/raygun4node/issues/216

TheRealAgentK:
> > No, that's different, isn't it. [breadcrumbs] is the tag for the debug module and that can be quite generic. Express adds something to it for instance that is not related to the file at all. I added the filename as an additional information for folks to easily find the file a debug statement from within the provider comes from.
> 
> The output looks correct to me:
> 
> ```
> 2024-05-14T12:21:08.984Z raygun [raygun.ts] Client initialized
> 2024-05-14T12:21:08.989Z raygun [raygun.breadcrumbs.ts] enter with new store
> ```
> 
> Nevertheless, I will change it to follow the rest of the files.

Yeah that output does look like I'd expect it to be.

TheRealAgentK:
> Did some more cleanup, updated README with Breadcrumb documentation, and applied many of the suggestions from Kai.
> 
> I mark this now as ready for review, I think it is in a state which is usable and good enough to go as first implementation.

I just gave it another look and I think it's good and my concerns/remarks are all addressed!

TheRealAgentK (approved):
LGTM

miquelbeltran:
Thanks for the review!

I'll go ahead and merge to develop, if we want to do any changes, we can do it before the release in a new PR

lib/breadcrumbs.ts
sumitramanga: @miquelbeltran Maybe open an issue on this?

README.md
```diff
@@ -291,6 +291,60 @@ const myBeforeSend = function (payload, exception, customData, request, tags) {
...
+
+```js
+// Add the Raygun Breadcrumb ExpressJS handler
+app.use(raygunClient.expressHandlerBreadcrumbs);
+
+// Setup the rest of the app, e.g.
+app.use("/", routes);
+```
+
+This middleware can be user together with the provided ExpressJS error handler `expressHandler`.
```
sumitramanga, line 340: "used" instead of "user"

README.md
```diff
@@ -291,6 +291,60 @@ const myBeforeSend = function (payload, exception, customData, request, tags) {
...
+The order in which the middlewares are configured is important. `expressHandlerBreadcrumbs` should go first to scope breadcrumbs correctly.
```
sumitramanga, line 341: Might be good to note above that this is only supported for Express?

examples/express-sample/app.js
```diff
@@ -34,6 +20,9 @@ raygunClient.user = function (req) {
 app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "ejs");
 
+// Add the Raygun breadcrumb Express handler
+app.use(raygunClient.expressHandlerBreadcrumbs);
+
 // uncomment after placing your favicon in /public
 // app.use(favicon(__dirname + '/public/favicon.ico'));
```
sumitramanga, line 27: Is this needed?

lib/raygun.breadcrumbs.ts
```diff
@@ -0,0 +1,139 @@
...
+    ...(breadcrumb as BreadcrumbMessage),
+    category: breadcrumb.category || "",
+    message: breadcrumb.message || "",
+    level: breadcrumb.level || "info",
+    timestamp: Number(new Date()),
+    type,
+    className: callsite?.fileName,
+    methodName: callsite?.functionName,
+
+    // TODO - do we need to do any source mapping?
```
sumitramanga, line 92: Create issue for this?

lib/raygun.ts
```diff
@@ -24,8 +25,10 @@ import {
   Tag,
   Transport,
 } from "./types";
+import * as breadcrumbs from "./raygun.breadcrumbs";
+import * as breadcrumbsExpressJs from "./raygun.breadcrumbs.express";
```
sumitramanga, line 29: If there's not many functions in here, we should import it directly instead with *

lib/raygun.breadcrumbs.express.ts
```diff
@@ -0,0 +1,34 @@
...
+  const crumbs = getBreadcrumbs();
+
+  if (!crumbs) {
+    debug(
+      "[raygun.breadcrumbs.express.ts] Add request breadcrumb skip, no store!",
+    );
+    return;
+  }
+
+  const internalCrumb: Breadcrumb = {
```
sumitramanga, line 21: Perhaps we need to add additional code comments around the internal breadcrumbs approach to mention what is user facing and internal

examples/express-sample/app.js
```diff
@@ -58,7 +47,9 @@ app.use(express.static(path.join(__dirname, "public")));
 app.use("/", routes);
 app.use("/users", users);
 
-// Add the Raygun Express handler
+// Add the Raygun error Express handler
 app.use(raygunClient.expressHandler);
```
sumitramanga, line 51: I agree that it (express naming usage)  is too tightly coupled at the moment. But until we see popular requests for other frameworks, I'm happy with the current implementation until we implement support for other frameworks. I think the new approach looks good (moving to `raygun.breadcrumbs.express.ts`).

But does a standalone node app need breadcrumbs tracked manually?

miquelbeltran: > But does a standalone node app need breadcrumbs tracked manually?

Yes, users would need to manually add breadcrumbs.

The `expressBreadcrumbsHandler` only adds a single breadcrumb for requests, so it isn't a big difference nevertheless.

examples/express-sample/app.js
miquelbeltran: The favicon was missing in the /public folder, this should be cleaned up, or we need a favicon. I'll see if we can get a raygun one.

lib/raygun.breadcrumbs.ts
miquelbeltran: Done: https://github.com/MindscapeHQ/raygun4node/issues/224

miquelbeltran:
@sumitramanga I will submit the suggestions in a new PR

lib/breadcrumbs.ts
miquelbeltran: Done https://github.com/MindscapeHQ/raygun4node/issues/224

Issue #138, opened 2024-01-25
Fix warning "module.parent.require is not supported by webpack"


tansongyang:
I have a repro of this issue at https://github.com/tansongyang/raygun-webpack-issue-repro.

Run `npm run build` in the console. Here's the output:

```
> next build

   ▲ Next.js 14.1.0

   Creating an optimized production build ...
 ⚠ Compiled with warnings

./node_modules/raygun/build/raygun.js
module.parent.require is not supported by webpack.

Import trace for requested module:
./node_modules/raygun/build/raygun.js
./app/raygun.ts
./app/page.tsx

./node_modules/raygun/build/raygun.js
Module not found: Can't resolve 'raygun-apm/lib/src/crash_reporting' in '/Users/frank/Code/raygun-webpack-issue-repro/node_modules/raygun/build'

Import trace for requested module:
./node_modules/raygun/build/raygun.js
./app/raygun.ts
./app/page.tsx
```

Source appears to be here: https://github.com/MindscapeHQ/raygun4node/blob/v0.13.0/lib/raygun.ts#L51

Not sure what the fix is, but if I navigate to the file in VS Code, I get a warning that `module.parent` is deprecated:

![image](https://github.com/MindscapeHQ/raygun4node/assets/9488719/4fb521c2-00dd-46fd-a090-c48b76fa92b9)

rob-mcgrail:
We've got the same issue

miquelbeltran:
Hello, sorry we took so long to respond here. I am now taking care of keeping this project up to date.
Thanks a lot for the sample project with the issue, that will help me debug and fix.
I'll see what I can do today.

miquelbeltran:
After looking into it, the main issue is how the package attempts to load the `raygun-apm` dependency.
As it tries to load it, and if fails. This is wrapped by a try-catch block, but still causes warnings to appear.

This causes two warnings when building the app:

```
1. module.parent.require is not supported by webpack.
2. Module not found: Can't resolve 'raygun-apm/lib/src/crash_reporting' in '...
```

I think a proper solution would be to change the way we provide the `apmBridge` internally, instead of trying to attempt to load it, I think we should expose a setter and let users provide it externally rather than attempt to load in internally.

I'll have to consult with the rest of the team on how to move forward with this.

miquelbeltran:
With the changes in the PR #222 the build output should be:

```
➜  raygun-webpack-issue-repro git:(main) ✗ npm run build

[...]

PR #222, opened 2024-05-16
fix: #138 New APM Bridge Setup

miquelbeltran:
### Description :memo:
- **Purpose**: The code that attempts to load the APM Bridge causes warnings on webpack users
- **Approach**: Extract the dependency, and perform setup on client code instead of internally.

As commented on the original ticket: https://github.com/MindscapeHQ/raygun4node/issues/138#issuecomment-2114320321

> The main issue is how the package attempts to load the raygun-apm dependency.
> As it tries to load it, and if fails. This is wrapped by a try-catch block, but still causes warnings to appear.
> 
> This causes two warnings when building the app:
> 
> 1. module.parent.require is not supported by webpack.
> 2. Module not found: Can't resolve 'raygun-apm/lib/src/crash_reporting' in '...
> 
> I think a proper solution would be to change the way we provide the apmBridge internally, instead of trying to attempt to load it, I think we should expose a setter and let users provide it externally rather than attempt to load in internally.

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Updates**

- Added a `setApmBridge` method.
- Added a new example project showing how to set up Raygun APM with Raygun4Node.
- Updated documentation.
- Added tests that verifies that the `apmBridge` is called.

### Test plan :test_tube:

- Verified that the module loading works using the example project.
- Verified that the tests run as expected.
- Verified that the build warning is gone using sample project in #138 
- enabled debug with `DEBUG=raygun-apm`, obtained the following logs from the Raygun APM package:

```
 raygun-apm [features] using feature level stable +0ms
  raygun-apm [features] RAYGUN_APM_CRASH_REPORTING_INTEGRATION -> enabled (level=stable) +53ms
  raygun-apm [crash-reporting] notified of error: Sending error with APM bridge 4be72872-7d32-403f-8d41-e6a037242fab +2ms
  raygun-apm [effects] exception {
  occurredAt: 14918346725n,
  errorClass: 'Sending error with APM bridge',
  correlationId: '4be72872-7d32-403f-8d41-e6a037242fab',
  triggerAsyncId: 1,
  threadId: 1
} +0ms
```

Verifying that the Raygun APM is called correctly.

- [x] Self-/dev-tested 


lib/raygun.ts
```diff
@@ -44,20 +45,7 @@ type SendOptionsResult =
...
-  | undefined
-  | null
-  | { notify(e: string | Error, s: string): void } = undefined;
-
-try {
-  if (module.parent) {
-    apmBridge = module.parent.require("raygun-apm/lib/src/crash_reporting");
-  } else {
-    apmBridge = require("raygun-apm/lib/src/crash_reporting");
-  }
```
miquelbeltran, line 57: The issue with the old code is that `module.parent` is deprecated: https://nodejs.org/api/deprecations.html#DEP0144

Then, the code attempts to load the module, but throws an error that appears as a build warning in webpack apps, e.g. `Module not found: Can't resolve 'raygun-apm/lib/src/crash_reporting' in ...`

That's why I think it is better to simply move this dependency outside the project and let users connect the two together, which requires an extra call.

examples/apm-sample/config/default.json
sumitramanga, line 1: The indentation looks to be quite large in this file. Is that intentional?

examples/apm-sample/config/production.json
sumitramanga, line 1: Same as above - The indentation looks to be quite large in this file. Is that intentional?

examples/apm-sample/README.md
```diff
@@ -0,0 +1,23 @@
+# Raygun Crash Reporting + APM Sample
+
+This is a sample application to show how to use Raygun4Node and Raygun-APM together.
+
+This example uses the local `raygun4node` package in the project root directory by simply pointing to the root directory as a dependency in package.json:
```
sumitramanga, line 5: It would be good to add what feature of Raygun we further see light up and if other things are required for set up. E.g. do they have to follow the Raygun APM for Node instructions. I don't know if this provider is required to function the APM tool?

examples/apm-sample/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "apm-sample",
+  "version": "1.0.0",
+  "main": "app.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "MIT",
+  "description": "",
```
sumitramanga, line 10: Would be good to fill this in?

lib/raygun.ts
```diff
@@ -24,6 +24,7 @@ import {
   SendOptions,
   Tag,
   Transport,
+  ApmBridge,
```
sumitramanga, line 27: Is there much of a performance hit now that this lives outside of the the project? 

I think this is much more cleaner than the previous implementation but just being wary of not having too much abstraction to avoid confusion :)

sumitramanga (changes requested):
Just a few comments, otherwise it's looking good!! Great job 💯

lib/raygun.ts
miquelbeltran: > Is there much of a performance hit now that this lives outside of the the project?

I think on the contrary. The previous code tried to load the APM package, failed, caught an exception, and ignored it. That's unnecessary work that happens every time the Raygun4Node package would start.

examples/apm-sample/README.md
miquelbeltran: There is some documentation in the root README.md as well. I think it would be better to reference that rather than duplicating the text.

examples/apm-sample/config/default.json
miquelbeltran: Indentation is the same as the other examples, the files were copied over.
I'd suggest doing a new PR, fixing the indentation of the config files for the three examples.

examples/apm-sample/package.json
miquelbeltran: Updated, cleaned up the package.json

miquelbeltran:
@sumitramanga I replied your questions, and simplified the `package.json`, so no need to include a description.

TheRealAgentK (approved):
This looks like a much cleaner approach, yes!

lib/raygun.ts
sumitramanga: Ahh okay, this is all good then !

sumitramanga (approved):
Always learning for you. Thank you for the changes. LGTM :shipit:

Issue #217, opened 2024-05-15
API: Allow filtering errors


miquelbeltran:
**Please describe your new feature request**

From comment: https://github.com/MindscapeHQ/raygun4node/issues/119#issuecomment-963938131

> Filtering out of errors/events before they hit your servers and count as an event. We have a lot of client side errors that are irrelevant and shouldn't be reported to us as application errors. Same on server side where we occasionally get errors in integrations which we don't need to report on.

**Describe the solution you'd like**

The Flutter SDK, like other implementations, provides the `onBeforeSend`.

This method allows you to either modify the outgoing message, or to discard it.

A similar solution could be provided for the Node implementation

**Describe alternatives you've considered**

- None yet

**Additional context**

See https://github.com/MindscapeHQ/raygun4flutter/blob/181f1871e6d27492e121154a65c5b8839fc72dba/lib/raygun4flutter.dart#L196

PR #231, opened 2024-05-20
fix: #217 allow skiping sending messages with `onBeforeSend()`

miquelbeltran:
### Description :memo:
- **Purpose**: Allow users to skip sending messages to Raygun
- **Approach**: Implement functionality in the `onBeforeSend()` method.

**Type of change**

**Updates**

According to the original `README.md`, in `onBeforeSend()` users could return `false` to skip sending the message.

This didn't work as expected, as the error was never skipped, and was sent to Raygun.

To fix this method, I allow users to return `null` instead of the `Message` to send.

For example, as in this example:

```js
raygunClient.onBeforeSend((message, exception, customData, request, tags) => {
  // If the message contains the word "skip", do not send the message to Raygun
  if (message.details.error.message.indexOf("skip") > 0) {
    return null; //skip sending
  }

  return message; //send original
});
```

For this, I modified the method type to allow returning `null`,  updated the functions that use it, and created tests and examples to verify it works as expected.

### Test plan :test_tube:

- Created unit tests.
- Added example in "express-sample"

- [ ] Self-/dev-tested 


miquelbeltran:
We can merge this after 1.0.0 has been released.

README.md
```diff
@@ -285,18 +285,18 @@ Call `Raygun.onBeforeSend()`, passing in a function which takes up to 5 paramete
...
+If, after inspecting the payload, you wish to discard it and abort sending it to Raygun, simply return `null`.
 
-By example:
+For example:
 
 ```javascript
 const myBeforeSend = function (payload, exception, customData, request, tags) {
   console.log(payload); // Modify the payload here if necessary
-  return payload; // Return false here to abort the send
+  return payload; // Return null here to abort the send
```
sumitramanga, line 299: Is this more clear - 
```suggestion
  return payload; // Return null here instead of payload to abort the send
```

examples/express-sample/app.js
```diff
@@ -50,6 +50,21 @@ app.use("/users", users);
 // Add the Raygun error Express handler
 app.use(raygunClient.expressHandler);
 
+// Optional: Configure onBeforeSend
+raygunClient.onBeforeSend((message, exception, customData, request, tags) => {
+  console.log(
+    `[app.js] onBeforeSend called with error: ${message.details.error.message}`,
+  );
+
+  // If the message contains the word "skip", do not send the message to Raygun
```
sumitramanga, line 59: This is a clear example, thank you!

examples/express-sample/views/index.ejs
```diff
@@ -11,6 +11,9 @@
     <div>
       <a href="/send">Send a custom error</a>
     </div>
+    <div>
+      <a href="/skip">Send an error that will be skipped</a>
```
sumitramanga, line 15: Maybe this instead - 
```suggestion
      <a href="/skip">Send an error that uses onBeforeSend and doesn't send to Raygun</a>
```

examples/express-sample/routes/index.js
```diff
@@ -9,6 +9,23 @@ router.get("/", function (req, res, next) {
...
+    .send("Error to skip")
+    .then((message) => {
+      res.render("send", {
+        title: "Skip sending message",
+        body: `Message should be null: ${message}`,
+      });
+    })
+    .catch((error) => {
+      res.render("send", {
+        title: "Failed to send error to Raygun",
```
sumitramanga, line 23: Should this say  -
```suggestion
        title: "Failed to skip sending error to Raygun",
```

sumitramanga (commented):
Good find! I've added some comments to cover please. Overall implementation looks good!

examples/express-sample/views/index.ejs
miquelbeltran: That won't be correct, because `onBeforeSend` is used by all attempts to send an error report.

examples/express-sample/routes/index.js
miquelbeltran: I don't think that would be correct, failing to skip would also mean that it was sent by mistake. Maybe just "There was an error" would be more accurate.

examples/express-sample/views/index.ejs
sumitramanga: I think I read this wrong! Will resolve this thread

examples/express-sample/routes/index.js
sumitramanga: Same as above. I think I read this wrong! Will resolve this thread

sumitramanga (approved):
LGTM woohoooo 💯

Issue #68, opened 2018-06-25
Can't Handle Symbol() Attribute On Error Object


ki4jnq:
We attach `Symbol()` type attributes to our known error conditions in our Express application. When Raygun tries to serialize these attributes the following error is raised:
```
TypeError: Cannot convert a Symbol value to a string
    at toHumanString (/Users/myusername/Projects/myProjectName/node_modules/object-to-human-string/lib/index.js:29:63)
    at RaygunMessageBuilder.setErrorDetails (/Users/myusername/Projects/myProjectName/node_modules/raygun/lib/raygun.messageBuilder.js:54:15)
    at Raygun.raygun.send (/Users/myusername/Projects/myProjectName/node_modules/raygun/lib/raygun.js:98:14)
    at Raygun.raygun.expressHandler (/Users/myusername/Projects/myProjectName/node_modules/raygun/lib/raygun.js:144:16)
    at WebFrameworkShim.applySegment (/Users/myusername/Projects/myProjectName/node_modules/newrelic/lib/shim/shim.js:1332:17)
    at _applyRecorderSegment (/Users/myusername/Projects/myProjectName/node_modules/newrelic/lib/shim/shim.js:949:20)
    at _doRecord (/Users/myusername/Projects/myProjectName/node_modules/newrelic/lib/shim/shim.js:928:17)
    at /Users/myusername/Projects/myProjectName/node_modules/newrelic/lib/shim/shim.js:913:24
    at Layer.handle_error (/Users/myusername/Projects/myProjectName/node_modules/express/lib/router/layer.js:71:5)
    at trim_prefix (/Users/myusername/Projects/myProjectName/node_modules/express/lib/router/index.js:315:13)
```

It appears that `object-to-human-string` tries to concatenate strings using the `+` operator, which specifically throws the above type error. It seems to me that attaching `Symbol()` metadata to error objects should not cause an error to be thrown.

Thanks for a great tool!

UberMouse:
Hi @ki4jnq,

We will see what we can do to actually fix this and get back to you. But for now, a workaround might be to set `useHumanStringForObject` to `false` in the options object passed to init.

Regards,
Taylor

miquelbeltran:
The issue seems to still be present, the `object-to-human-string` package didn't really have a new release for 9 years.

I was able to reproduce this in unit tests:

```js
    var builder = new MessageBuilder({ useHumanStringForObject: true });
[...]

PR #255, opened 2024-06-14
fix: #68 Better humanize string functionality

miquelbeltran:
### Description :memo:
- **Purpose**: Fixes issue #68 where classes like `Symbol` caused crashes
- **Approach**: Implements a custom humanize method, similar to the original one.

The method has two key advantages:

- Generates a String as a single line
- Goes only 1 level deep into object properties.

The advantage over using `JSON.stringify` is that we have better control over how the String is constructed.

We don't need to worry on recursive/self-referencing objects since only one level is explored.

The new method preserves a similar format to the original one, so it should not disrupt current users.

**Type of change**

**Updates**

- Removed dependency `object-to-human-string` (which has not been updated for 9 years!)
- Implemented humanize error method.
- Added more tests

### Test plan :test_tube:

- Unit tests.
- Should test with examples as well.

- [ ] Self-/dev-tested 

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
   "dependencies": {
     "@types/express": "^4.17.21",
     "debug": "^4.3.4",
-    "object-to-human-string": "0.0.3",
     "stack-trace": "0.0.10",
     "uuid": "^9.0.1"
   },
--- a/examples/express-sample/package-lock.json
+++ b/examples/express-sample/package-lock.json
@@ -22,7 +22,7 @@
     },
     "../..": {
       "name": "raygun",
-      "version": "0.15.0-0",
+      "version": "1.1.0",
       "dependencies": {
         "@types/express": "^4.17.21",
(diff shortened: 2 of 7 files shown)


miquelbeltran:
Note: To merge after 1.1.0 is released

package.json
```diff
@@ -63,7 +63,6 @@
   "dependencies": {
     "@types/express": "^4.17.21",
     "debug": "^4.3.4",
-    "object-to-human-string": "0.0.3",
```
sumitramanga, line 66: Awesome drop 💪🏽 9 years is a long time!

sumitramanga (approved):
LGTM 🧊

sumitramanga:
I'll test in the `develop` branch

TheRealAgentK (approved):
LGTM

miquelbeltran:
Thanks! merging now

PR #263, opened 2024-06-21
feat: #216 register console.log calls to breadcrumbs

### LINKED ISSUE #216: Breadcrumbs: Capture console log messages as breadcrumbs
**Please describe your new feature request**

The idea is to setup a way to capture calls to `console.info` etc. as Breadcrumbs.

**Describe the solution you'd like**

A possible solution was found in the original `breadcrumbs` feature branch:

```js
 const consoleMethods: [keyof typeof console, InternalBreadcrumb["level"]][] = [
   ["debug", "debug"],
   ["log", "info"],
   ["info", "info"],
   ["warn", "warning"],
   ["error", "error"],
 ];

 for (const [method, level] of consoleMethods) {
   const oldMethod = (console as any)[method];
   (console as any)[method] = function logWithBreadcrumb<T>(
     this: T,
     ...args: any[]
   ) {
     addBreadcrumb({ message: args.join(" "), level }, "console");
     return oldMethod.apply(this, args);
   };
 }
```

**Describe alternatives you've considered**

- TBD

**Additional context**

Original comment suggestion: https://github.com/MindscapeHQ/raygun4node/pull/210#discussion_r1600961119

miquelbeltran:
### Description :memo:
- **Purpose**: Implements suggestion in #216 
- **Approach**: Extends the console methods which will register breadcrumbs automatically
- Closes #216 

**Type of change**

**Updates**

- Add configuration param `consoleBreadcrumbs`.
- Implemented `setupConsoleBreadcrumbs` method
- Created test

### Test plan :test_tube:

- Added `console.log` and similar to `express-sample` and tested.

The screenshot show the console breadcrumbs:

Screenshot from 2024-06-21 11-53-54

In this screenshot, the "[sass] skip: ...." is added by Express itself internally, the "Call /send" is added by us. The icon on the left shows that it is a "console" type breadcrumb.

- [x] Self-/dev-tested 

### DIFF
--- a/lib/raygun.breadcrumbs.ts
+++ b/lib/raygun.breadcrumbs.ts
@@ -1,7 +1,9 @@
 import type { AsyncLocalStorage } from "async_hooks";
 import type { BreadcrumbMessage, Breadcrumb } from "./types";
 import { format } from "node:util";
+import ansiRegex from "ansi-regex";
 
+const regex = ansiRegex();
 const debug = require("debug")("raygun");
 
 let asyncLocalStorage: AsyncLocalStorage<Breadcrumb[]> | null = null;
@@ -175,13 +177,9 @@ export function setupConsoleBreadcrumbs() {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ...args: any[]
     ) {
-      // Dynamically import module because require is not supported
-      import("ansi-regex").then((ansiRegex) => {
-        const regex = ansiRegex.default();
-        // Call to node-util format like the original console methods
-        // and remove any ansi encoding (e.g. console colors) using replace
-        addBreadcrumb({ message: format(...args).replace(regex, ""), level }, "console");
-      });
+      // Call to node-util format like the original console methods
+      // and remove any ansi encoding (e.g. console colors) using replace
+      addBreadcrumb({ message: format(...args).replace(regex, ""), level }, "console");
       // Still call original method
       return oldMethod.apply(this, args);
     };
--- a/lib/raygun.breadcrumbs.ts
+++ b/lib/raygun.breadcrumbs.ts
@@ -179,7 +179,10 @@ export function setupConsoleBreadcrumbs() {
     ) {
       // Call to node-util format like the original console methods
       // and remove any ansi encoding (e.g. console colors) using replace
-      addBreadcrumb({ message: format(...args).replace(regex, ""), level }, "console");
+      addBreadcrumb(
+        { message: format(...args).replace(regex, ""), level },
+        "console",
+      );
       // Still call original method
       return oldMethod.apply(this, args);
     };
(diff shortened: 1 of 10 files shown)


lib/raygun.breadcrumbs.ts
```diff
@@ -151,3 +155,36 @@ export function clear() {
...
+  // Map console methods to a Breadcrumb level
+  const consoleMethods: [keyof typeof console, Breadcrumb["level"]][] = [
+    ["debug", "debug"],
+    ["log", "info"],
+    ["info", "info"],
+    ["warn", "warning"],
+    ["error", "error"],
+  ];
+
+  // Extend all console methods to call to addBreadcrumb
```
TheRealAgentK, line 169: I haven't run this PR yet, but what happens with the internal (as in: internal to our provider) calls of `console.<something>` ? Wouldn't those trigger this logic as well?

miquelbeltran: If the user configured `setupConsoleBreadcrumbs` to `true` then yes, all calls to `console.log` get registered as breadcrumbs as well, not just the ones the user does, but _all of them_ including the ones in 3rd party packages.

TheRealAgentK: >If the user configured setupConsoleBreadcrumbs to true then yes, all calls to console.log get registered as breadcrumbs as well, not just the ones the user does, but all of them including the ones in 3rd party packages.

Hm, when I suggested trying that in the old PR comment I wasn't really clear of this side effect. I'm not sure that this is such a good feature then. It seems it'll lead to a lot of stuff being logged as breadcrumbs that users would quite possibly not want to be logged in there. Very noisy and cluttered as a result.

Should we pull this PR from 1.2.0 and have a further though on the feature itself?

@sumitramanga Any opinions on this?

sumitramanga: Yeah I agree with Kai. I can check out what the other providers are doing and see if this is the correct behaviour to have

miquelbeltran: Seems that the `raygun4js` also does it: https://github.com/MindscapeHQ/raygun4js/blob/2d0eeb333b1c3e28df26c9597e78b034b30afb91/src/raygun.breadcrumbs.js#L207

I would be surprised if any other implementation besides JavaScript does it, since it is fairly hacky to override the original `console.log` methods, and I am not sure other platforms may support it.

Otherwise, the feature is disabled by default, so it won't affect users unless they enable it.

TheRealAgentK: I acknowledge all of your points - it's disabled by default and it's the same in RG4JS.

But: I'm not sure that - given the side-effects when used - this is actually a good feature in the first place. I mean, sure we can have it and it's implemented as well as it can be done, but the feature feels a bit meh now and almost not good practice to even offer it.

Issue #218, opened 2024-05-15
API: Review setUser method


miquelbeltran:
**Please describe your new feature request**

From the original comment: https://github.com/MindscapeHQ/raygun4node/issues/119#issuecomment-963938131

> Setting the user with a function is cumbersome and impractical, as we don't always run in the context of a "request", but sometimes run in a worker or task context.

**Describe the solution you'd like**

Flutter and Android implementations support both a method to set the user ID `setUserId(String)` and a method to set the user data `setUser(UserInfo)`.

Node implementation offers:

```js
  // This function is deprecated, is provided for legacy apps and will be
  // removed in 1.0: use raygun.user instead
  setUser(user: RawUserData) {
```

However, it is marked as deprecated.

I think we should keep it around

**Describe alternatives you've considered**

Having a `setUserId` method could be useful?

**Additional context**

See: https://github.com/MindscapeHQ/raygun4flutter/blob/181f1871e6d27492e121154a65c5b8839fc72dba/lib/raygun4flutter.dart#L148-L178

PR #269, opened 2024-06-24
feat: #218 Optional `userInfo` in `send()`

miquelbeltran:
### Description :memo:
- **Purpose**: Improve how users can provide user information to their error reports
- **Approach**: 

Users have previously complained that the current implementation, based on the request, cannot be used in all situations. (see comments in #218).

The other older way, was to call to a global `setUser(user)` method, but that will cause problems when handling multiple requests by the same client, as not all requests are linked to the same user.

To fix that, I take inspiration on how the `raygun4net` package does it: https://github.com/MindscapeHQ/raygun4net/blob/ff5ef7eb6a3adaa7fdd49f2cd835fe8e68aef63d/Mindscape.Raygun4Net4/RaygunClient.cs#L339

By allowing users to pass the user information in the `send()` method, they can customize the way they provide the user information and ensure that the correct user information is attached to the error report.

e.g.

```js
client.send(new Error(), { userInfo });
```

This user information will take priority over the currently existing methods.

But, at the same time, the raygun client still works as before, so users don't need to change their code.

- Closes: #218 

**Type of change**

**Updates**

- Add optional `userInfo` parameter to `send()`.
- Pass user information to the internal message builder.
- Add tests.

### Test plan :test_tube:

- Unit test implemented.
- Implemented custom userInfo in express example:

Example, adding this user information:

```js
// Provide custom user info in /send endpoint
  const userInfo = {
    identifier: "123456",
    email: "custom@example.com",
  };

  raygunClient
    .send("Custom Raygun Error in /send endpoint", { tags: ["request"], userInfo })
```

Gets reported as:

image

- [x] Self-/dev-tested 

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,12 +237,11 @@ This package offers three different ways to do that:
 
 The following properties can be provided as user information:
 
-- `identifier` is the user identifier. This will be used to uniquely identify the user within Raygun. This is the only required parameter, but is only required if you are using customers tracking.
-- `uuid` is the identifier of the device the app is running on. This could be used to correlate user accounts over multiple machines.
-- `firstName` is the user's first or preferred name.
-- `lastName` is the user's surname name.
-- `fullName` is the user's full name.
-- `email` is the user's email address.
+- `identifier`: Unique identifier for the user is the user identifier.
+- `email`: User's email address.
+- `firstName`: User's first name (what you would use if you were emailing them - "Hi {{firstName}}, ...")
+- `fullName`: User's full name.
+- `uuid`: Device unique identifier. Useful if sending errors from a mobile device.
 
 All properties are `strings`. Any other properties will be discarded.
 
@@ -252,10 +251,16 @@ Example:
 userInfo = {
     identifier: "123",
     email: "user@example.com",
-    fullName: "Fullname",
     firstName: "First name",
+    fullName: "Fullname",
     uuid: "a25dfe58-8db3-496c-8768-375595139375",
 }
+```
+
+For legacy support reasons, you can also provide the `string` identifier directly as the user information:
+
+```js
+setUser("123");
 ```
 
 #### `userInfo` parameter in `send()`
--- a/README.md
+++ b/README.md
@@ -238,11 +238,11 @@ This package offers three different ways to do that:
 The following properties can be provided as user information:
 
 - `identifier` is the user identifier. This will be used to uniquely identify the user within Raygun. This is the only required parameter, but is only required if you are using customers tracking.
-- `isAnonymous` is a bool indicating whether the user is anonymous or actually has a user account. Even if this is set to true, you should still give the user a unique identifier of some kind.
-- `email` is the user's email address.
-- `fullName` is the user's full name.
-- `firstName` is the user's first or preferred name.
 - `uuid` is the identifier of the device the app is running on. This could be used to correlate user accounts over multiple machines.
+- `firstName` is the user's first or preferred name.
+- `lastName` is the user's surname name.
+- `fullName` is the user's full name.
+- `email` is the user's email address.
 
 All properties are `strings`. Any other properties will be discarded.
 
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ backwards compatibility; the Node-style callback should be preferred.
 The `send()` method accepts a series of optional named parameters, defined as follows:
 
 ```js
-client.send(error, { customData, request, tags });
+client.send(error, { customData, request, tags, timestamp, userInfo });
 ```
 
 Each one of these parameters is optional.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -222,7 +222,55 @@ client.send(new Error(), { timestamp: new Date(2024, 5, 13, 10, 0, 0) });
...
+#### User information object
+
+The following properties can be provided as user information:
+
+- `identifier` is the user identifier. This will be used to uniquely identify the user within Raygun. This is the only required parameter, but is only required if you are using customers tracking.
+- `isAnonymous` is a bool indicating whether the user is anonymous or actually has a user account. Even if this is set to true, you should still give the user a unique identifier of some kind.
+- `email` is the user's email address.
+- `fullName` is the user's full name.
+- `firstName` is the user's first or preferred name.
+- `uuid` is the identifier of the device the app is running on. This could be used to correlate user accounts over multiple machines.
```
TheRealAgentK, line 245: I'm not sure if this should be tracked with the user. In other providers we're not doing this.  To me, the thing that ties users together is `identifier` and if they come from different devices, should this not be set somewhere/somehow differently?

README.md
```diff
@@ -222,7 +222,55 @@ client.send(new Error(), { timestamp: new Date(2024, 5, 13, 10, 0, 0) });
...
+
+All properties are `strings`. Any other properties will be discarded.
```
TheRealAgentK, line 247: `isAnonymous` should be a boolean, I think.

TheRealAgentK (commented):
I think the change itself is fine - I'm not sure though what exactly the situation is re `isAnonymous`. Did a quick search in the repo and couldn't find how that's derived/treated in general (even outside of this PR), but I might have missed the right place.

README.md
miquelbeltran: No idea, that's how it worked before. This PR didn't change the UserInfo object.

README.md
miquelbeltran: Seems that this property doesn't even exist in the actual user info object, I'll clean up the documentation.

image

miquelbeltran:
The Raygun API spec shows the following properties:

```json
// Information about the user that caused the error
		"user": {
			// Unique identifier for the user
			"identifier": "123456789",
			// Flag indicating if the user is anonymous or not
			"isAnonymous": false,
			// User's email address
			"email": "test@example.com",
			// User's full name
			"fullName": "Test User",
			// User's first name (what you would use if you were emailing them - "Hi {{firstName}}, ...")
			"firstName": "Test",
			// Device unique identifier. Useful if sending errors from a mobile device.
			"uuid": "783491e1-d4a9-46bc-9fde-9b1dd9ef6c6e"
		},
```

But the implementation is as follows:

```js
export type RawUserData = {
  identifier?: string;
  uuid?: string;
  firstName?: string;
  lastName?: string;
  fullName?: string;
  email?: string;
};
```

I will align the Raygun4Node implementation to the spec in this PR.

README.md
miquelbeltran: Indeed, in the Flutter client, the `uuid` property is not used. Will ask on Slack for clarification.

miquelbeltran:
FYI @TheRealAgentK I did some changes in the implementation:

- Added the internal `isAnonymous` property. This aligns to how it works in the Flutter client (and I guess Android client as well)
- Aligned documentation with the Raygun API documentation.
- Cleanup the user code a bit, allowing users to call with an identifier directly `setUser("123")`, seems that this was already in place for legacy purposes but not properly implemented, so added clarification.

I will ask on Slack about the `uuid` property.

TheRealAgentK:
> FYI @TheRealAgentK I did some changes in the implementation:
> 
> * Added the internal `isAnonymous` property. This aligns to how it works in the Flutter client (and I guess Android client as well)
> * Aligned documentation with the Raygun API documentation.
> * Cleanup the user code a bit, allowing users to call with an identifier directly `setUser("123")`, seems that this was already in place for legacy purposes but not properly implemented, so added clarification.
> 
> I will ask on Slack about the `uuid` property.

So, the only remaining question I have is - are we even using/setting `isAnonymous` anywhere? 

I think it's ok for it to be on the object, but if not in this PR, then in a follow-up we should actually set it to `true`, `false` following some rules. If these rules are what I outlined on Slack or not (following the logic of RG4Android etc) can be a separate discussion I think.

Maybe the easiest way to move this along is a new issue for adding this logic and then we can get this PR over the line?

I'll approve it so that @sumitramanga can merge it and play with the feature for now.

TheRealAgentK (approved):

sumitramanga:
I'd say `isAnonymous` should be a way for people to manually set. Because it is an easy way to do so. Such as, if the customer's user decides, for sensitive reasons, they want to remain anonymous like an opt out of collecting my data, then there is a simple way to call this in our customers code.

If I am understanding correctly that `isAnonymous` is only ever used for internal purposes!

sumitramanga:
I'll still go ahead and merge. As Kai said, we can open up a new issue and work from there. I'm also happy releasing this as a feature as is!

PR #271, opened 2024-06-26
chore: Release 1.2.0

miquelbeltran:
### Description :memo:
- **Purpose**: Release 1.2.0
- **Approach**: Merge to `master`. DO NOT SQUASH!

- Depends on https://github.com/MindscapeHQ/raygun4node/pull/272

**Type of change**

- [x] Release

**Updates**

- See CHANGELOG

### Test plan :test_tube:

- Tested with local example project

- [x] Self-/dev-tested 

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,25 @@ Tags can also be set globally using setTags
 client.setTags(['Tag1', 'Tag2']);
 ```
 
+### Timestamp
+
+You can specify the exact time your error occurred in the Send() function with the `timestamp` parameter.
+Otherwise, the current time will be used.
+
+This can be useful when combining Raygun together with other logger tools that provide a timestamp.
+
+In milliseconds since epoch:
+
+```javascript
+client.send(new Error(), { timestamp: 1718268992929 });
+```
+
+As `Date` object:
+
+```javascript
+client.send(new Error(), { timestamp: new Date(2024, 5, 13, 10, 0, 0) });
+```
+
 ### Customers
 
 New in 0.4: You can set **raygunClient.user** to a function that returns the user name or email address of the currently logged in user.
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ backwards compatibility; the Node-style callback should be preferred.
 The `send()` method accepts a series of optional named parameters, defined as follows:
 
 ```js
-client.send(error, { customData, request, tags });
+client.send(error, { customData, request, tags, timestamp, userInfo });
 ```
 
 Each one of these parameters is optional.
@@ -222,7 +222,60 @@ client.send(new Error(), { timestamp: new Date(2024, 5, 13, 10, 0, 0) });
 
 ### Customers
 
-New in 0.4: You can set **raygunClient.user** to a function that returns the user name or email address of the currently logged in user.
+You can attach user information to every Raygun Crash Report.
+
+It will be transmitted with the error sent, and a count of affected customers will appear on the dashboard in the error group view.
+If you provide an email address, and the user has associated a Gravatar with it, their picture will be also displayed.
+
+This package offers three different ways to do that:
+
+1. Provide the `userInfo` parameter in the `send()` method.
+2. Implement the `user(request)` method.
+3. Call to the `setUser(user)` method (not recommended and deprecated).
+
+#### User information object
+
+The following properties can be provided as user information:
+
+- `identifier`: Unique identifier for the user is the user identifier.
+- `email`: User's email address.
+- `firstName`: User's first name (what you would use if you were emailing them - "Hi {{firstName}}, ...")
+- `fullName`: User's full name.
+- `uuid`: Device unique identifier. Useful if sending errors from a mobile device.
+
+All properties are `strings`. Any other properties will be discarded.
+
+Example:
+
+```js
+userInfo = {
+    identifier: "123",
+    email: "user@example.com",
+    firstName: "First name",
+    fullName: "Fullname",
+    uuid: "a25dfe58-8db3-496c-8768-375595139375",
+}
+```
+
+For legacy support reasons, you can also provide the `string` identifier directly as the user information:
+
+```js
+setUser("123");
+```
+
+#### `userInfo` parameter in `send()`
+
+Provide the `userInfo` optional parameter in the `send()` method call:
+
+```javascript
+client.send(new Error(), { userInfo });
+```
+
+This provided user information will take priority over the `user(request)` and `setUser(user)` methods.
+
+#### Implement `raygunClient.user(req)`
+
+You can set `raygunClient.user` to a function that returns the user name or email address of the currently logged in user.
 
 An example, using the Passport.js middleware:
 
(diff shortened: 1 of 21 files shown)


sumitramanga:
I'll look to review and test once the 2 remaining PR's are merged

sumitramanga (approved):
LGTM 💛

README.md
```diff
@@ -223,32 +295,15 @@ raygunClient.user = function (req) {
...
-This will be transmitted with each message sent, and a count of affected customers will appear on the dashboard in the error group view. If you return an email address, and the user has associated a Gravatar with it, their picture will be also displayed.
-
-If you return an object, it may have any of the following properties (only identifier is required):
-
-`identifier` is the user identifier. This will be used to uniquely identify the user within Raygun. This is the only required parameter, but is only required if you are using customers tracking.
-
-`isAnonymous` is a bool indicating whether the user is anonymous or actually has a user account. Even if this is set to true, you should still give the user a unique identifier of some kind.
-
-`email` is the user's email address.
+#### Global `setUser(user)` method
```
PanosNB, line 301: Should we add the deprecated note here at the method's title as well?

README.md
```diff
@@ -553,6 +609,7 @@ All requests use the same authentication methods as the upload call (Basic Authe
 ### Known Issues
 
 - Node will show compilation warnings when using Raygun4Node in Webpack applications.
+- Although Breadcrumbs report the source filename, code line and function name, these are not processed using source maps like with stack traces.
```
PanosNB, line 612: Do we plan to tackle this in an upcoming release?

lib/raygun.batch.ts
```diff
@@ -1,31 +1,33 @@
...
-  Callback,
-  HTTPOptions,
-  SendOptions,
-} from "./types";
+import { HTTPOptions, SendOptions } from "./types";
 
 const debug = require("debug")("raygun");
 
-export type MessageAndCallback = {
+export type MessageAndPromise = {
```
PanosNB, line 8: If `MessageAndCallback` was somehow surfaced to the user, we are introducing a breaking change here

lib/raygun.batch.ts
```diff
@@ -1,31 +1,33 @@
...
   messageSizeInBytes: number;
 };
 
+// Contains the resolve and reject methods of the original promise
+// Allows to call to resolve and reject from outside the Promise scope
+type PromiseReference<T> = {
+  resolve: (message: T) => void;
+  reject: (error: Error) => void;
+};
+
```
PanosNB, line 30: Should these constants be set to their default values in the raygun client initialization?

lib/raygun.batch.ts
```diff
@@ -52,24 +54,13 @@ export class RaygunBatchTransport {
...
-          }
-        },
-      });
+    const promise = this.onIncomingMessage(options.message);
 
-      if (!this.timerId) {
-        this.timerId = setTimeout(() => this.processBatch(), 1000);
-      }
-    });
+    if (!this.timerId) {
```
PanosNB, line 59: We don't need `=== undefined` because this is not changed anywhere else and setTimeout will never return 0, or false, right?

lib/raygun.human.ts
```diff
@@ -0,0 +1,54 @@
...
+    return "...";
+  }
+
+  let out = "";
+
+  // Iterate over all object properties
+  Object.keys(obj).forEach(function (key) {
+    // Recursive call increases level
+    out +=
+      symbolSeparator + key + symbolEqual + humanString(obj[key], level + 1);
```
PanosNB, line 42: I am afraid this suffers from the stack-overflow cyclical object problem, similarly to the last time :-)

For example:

```
const a = {};
const b = { a };
a.b = b;

console.log(humanString(a));
```

Instead, we need to keep track of the objects already visited in a set, and before invoking the recursive call, confirm they are not already visited

lib/raygun.messageBuilder.ts
```diff
@@ -137,7 +137,7 @@ export class RaygunMessageBuilder {
     this._filters = options.filters || [];
 
     this.message = {
-      occurredOn: new Date(),
+      occurredOn: options.timestamp || new Date(),
```
PanosNB, line 140: Let's hope no one actually meant 0 as a legit timestamp! :-)

test/raygun.messageBuilder_test.js
```diff
@@ -54,7 +54,46 @@ test("basic builder tests", function (t) {
...
     var message = builder.build();
     tt.ok(message.details.error.message);
-    tt.equal("name=Test", message.details.error.message);
+    tt.equal(message.details.error.message, "name=Test");
+    tt.ok(message.details.groupingKey);
+    tt.end();
+  });
+
+  // Reference: https://github.com/MindscapeHQ/raygun4node/issues/68
+  t.test("humanise Symbol object string", function (tt) {
```
PanosNB, line 63: Sorry, US English only please :-)

Humanize

PanosNB: Same as in other instances below

test/raygun.messageBuilder_test.js
```diff
@@ -54,7 +54,46 @@ test("basic builder tests", function (t) {
...
+    tt.ok(message.details.error.message);
+    tt.equal(message.details.error.message, "symbol=Symbol(Test)");
+    tt.ok(message.details.groupingKey);
+    tt.end();
+  });
+
+  t.test("humanise self-referencing complex error string", function (tt) {
+    var builder = new MessageBuilder({ useHumanStringForObject: true });
+
+    // error self-references, causing a potential infinite recursion in the filter method
```
PanosNB, line 76: Hmmm... I just argued above this should not be allowed. I see you did consider it but plan it as expected behavior.

I think we should not allow such a crash, instead, replace a cyclical object's value with something like `[Circular]`

PanosNB (changes requested):
Great work overall! A few questions and comments, but mainly protection from objects with cyclical references and doublecheck we haven't exposed that class we are changing its definition

README.md
miquelbeltran: No, this is not a client issue, but a general problem with Raygun that should be implemented on the server side. I thought about mentioning this here since the topic of source mapping and breadcrumbs came up in a comment.

lib/raygun.batch.ts
miquelbeltran: No, it is not.

lib/raygun.batch.ts
miquelbeltran: This is not a client facing API

lib/raygun.batch.ts
miquelbeltran: It is indeed initialized to null here: https://github.com/MindscapeHQ/raygun4node/blob/7d9a3cf677757ee6fe7e14eb3794e8f256bf6cee/lib/raygun.batch.ts#L36C11-L36C18

This PR did not change how the `timerId` works.

lib/raygun.human.ts
miquelbeltran: No, it is not because the algorithm only goes one level deep, as stated adobe.

lib/raygun.messageBuilder.ts
miquelbeltran: We checked that, the Raygun server ignores any values of `occurredOn` that are like 128 days apart from the real date.

test/raygun.messageBuilder_test.js
miquelbeltran: The code already does that. This is a unit test verifying it cannot happen.

The code replaces anything deeper than level 1 with `...` independently if it is a circular reference or not.

All of this works similar to the 3rd party library it replaced.

README.md
miquelbeltran: Will do that

test/raygun.messageBuilder_test.js
miquelbeltran: Will do!

miquelbeltran:
Changes submitted here https://github.com/MindscapeHQ/raygun4node/pull/275 @PanosNB

PanosNB (approved):
Good work Miguel!

sumitramanga (approved):

PR #320, opened 2024-09-29
feat: Add configurable timeouts 

### LINKED ISSUE #311: Question: any way to abort long-running `send` call?
This isn't a bug or feature request, but opening issue as a question. 

I'm noticing that occasionally there could be a while before Raygun's `send` call responds and holds up connections that can be used for other sources on my server. As a result, whenever there are many send calls, my server takes quite a hit and slows down. In these cases, is it possible to set a timeout for the send call?

### LINKED ISSUE #319: Transport: Timeouts missing
**Describe the bug**
Timeouts are missing in the transport layer. Ideally they should be configurable.

**Expected behavior**
The provider should have a default send timeout of 5 seconds. The timeout should be configurable though similar to other configurable values.

Refer to discussion in: #311

TheRealAgentK:
### Description :memo:
- Fixes #319 and possibly improve or fix #311 (to be seen)
- Adds `timeout` as an option and an explicit default HTTP timeout of 5000ms
- Changes transport layer behaviour to destroy the request after various lifecycle events, in particular when the transport promise was completed successfully. This works around the default behaviour of `http` and `https` modules in Node to not stop a request even when it runs into a timeout
- Unrelated change in the debug logs for breadcrumbs

**Type of change**

**Updates**

:point_right: Changed TS types
:point_right: Changes to transport layer
:point_right: Introduced new option
:point_right: Added tests to the mocking test bed
:point_right: TDB: Update documentation

### Test plan :test_tube:
The easiest way to test under RL scenarios is with the express sample app. Set an API key and then configure `raygun.client.js` with options such as for example:

```
var raygunClient = new raygun.Client().init({
  apiKey: config.Raygun.Key,
  timeout: 10000,
  batch: true,
  batchFrequency: 10000
});
```

- [x] Self-/dev-tested 

### DIFF
--- a/lib/raygun.ts
+++ b/lib/raygun.ts
@@ -148,7 +148,7 @@ class Raygun {
           port: this._port,
           useSSL: !!this._useSSL,
           apiKey: this._apiKey,
-          timeout: this._timeout || DEFAULT_TIMEOUT
+          timeout: this._timeout || DEFAULT_TIMEOUT,
         },
       });
     }
@@ -589,7 +589,7 @@ class Raygun {
       port: this._port,
       useSSL: this._useSSL || false,
       apiKey: this._apiKey || "",
-      timeout: this._timeout || DEFAULT_TIMEOUT, 
+      timeout: this._timeout || DEFAULT_TIMEOUT,
     };
 
     return {


lib/types.ts
miquelbeltran, line 1: Could be possible to unify `timeout` and `batchTimeout` on `RaygunOptions`? Since the code will use one or the other depending on if `batch` is true. Otherwise, it could be the case that a user sets `timeout` and then `batch` true, and the `timeout` isn't being used because it expects `batchTimeout` instead.

lib/raygun.ts
```diff
@@ -570,7 +576,8 @@ class Raygun {
           host: this._host,
           port: this._port,
           useSSL: !!this._useSSL,
-          apiKey,
+          apiKey: apiKey,
+          ...(!this._batch && { timeout: this._timeout || DEFAULT_TIMEOUT }), // This can be called via batch mode, where the normal timeout is not used, so only set if batch is disabled
```
miquelbeltran, line 580: I think that if the `RaygunOptions.timeout` and `RaygunOptions.batchTimeout` would be unified, it wouldn't be necessary to do this `!this._batch` check.

lib/raygun.transport.ts
```diff
@@ -57,12 +59,32 @@ export function send(
...
+          // if a timeout was set, destroy the request after successful completion
+          if (options.http.timeout) {
+            debug(
+              `[raygun.transport.ts] Destroying request for message: ${options.message}`,
+            );
+            request.destroy();
+            debug(
+              `[raygun.transport.ts] Request destroyed for message: ${options.message}`,
+            );
+          }
```
miquelbeltran, line 71: I think we should destroy the request also when `timeout` is not set.

miquelbeltran (commented):
Overall looks good! I think it can be simplified a bit if the RaygunOptions get unified to a single config property for timeout.

Also, we should document this option in the `README.md` before merging.

miquelbeltran (approved):
LGTM!

PR #452, opened 2026-01-27
fix: replace uuid with crypto.randomUUID() for CommonJS compatibility

### LINKED ISSUE #451: raygun@2.2.0 breaks CommonJS consumers due to ESM-only uuid dependency
###    Issue 

raygun@2.2.0 updated the uuid dependency to ^13.0.0 (or ^12.0.0+), which is now an ESM-only package. This is breaking on node versions < 22 and in AWS Lambda (where `--experimental-require-module` is simply not supported even in v22)

### Root Cause

The raygun package is compiled to CommonJS (using "module": "commonjs" in tsconfig.json). This results in `require("uuid")` calls in the distributed code.

`Error: require() of ES Module .../node_modules/uuid/dist-node/index.js from .../node_modules/raygun/build/raygun.js not supported.`

### Reproduction:
1. Create a CommonJS Node.js project
2. npm install raygun@2.2.0
3. require('raygun')
4. Run the script using node < 22 -> it will crash.

Note: make sure `--experimental-require-module` is not true when testing this, can check using `node --help | grep -i experimental | grep "is true"`

TheRealAgentK:
### Description :memo:
- **Purpose**: Fixes #451 
- **Approach**: Replaces `uuid` dependency with built-in `crypto.randomUUID()`

**Type of change**

**Updates**

:point_right: Removed `uuid`, replaced with `crypto.randomUUID()`
:point_right: Added script to test the broken / valid behaviour for CommonJS modules
:point_right: Documented the change

- [x] Self-/dev-tested 

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "main": "build/raygun.js",
   "types": "build/raygun.d.ts",
   "engines": {
-    "node": ">= 0.10.0"
+    "node": ">= 14.17.0"
   },
   "files": [
     "build/",
@@ -40,7 +40,9 @@
     "//prepare": "prepare project for distribution",
     "prepare": "tsc",
     "//test": "runs tests on main project",
-    "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js"
+    "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js",
+    "//test:cjs": "tests CommonJS compatibility (catches ESM-only deps)",
+    "test:cjs": "node --no-experimental-require-module scripts/test-cjs-compat.js"
   },
   "devDependencies": {
     "@eslint/js": "^9.2.0",
@@ -48,7 +50,6 @@
     "@stylistic/eslint-plugin-ts": "^4.1.0",
     "@types/node": "^24.0.8",
     "@types/stack-trace": "0.0.33",
-    "@types/uuid": "^10.0.0",
     "eslint": "^9.16.0",
     "eslint-plugin-tsdoc": "^0.5.0",
     "express": "^5.1.0",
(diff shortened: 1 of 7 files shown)


amurf:
Thank you for the quick turn around.

A small thing but I noticed you have:

```
  "engines": {
    "node": ">= 0.10.0"
  }
```
in package.json

I believe this method is only supported since `14.17.0` so would be a good idea to update that at the same time.

TheRealAgentK:
> Thank you for the quick turn around.
> 
> A small thing but I noticed you have:
> 
> ```
>   "engines": {
>     "node": ">= 0.10.0"
>   }
> ```
> 
> in package.json
> 
> I believe this method is only supported since `14.17.0` so would be a good idea to update that at the same time.

Yeah, I saw that when I made the fix and changed it:

```
  "engines": {
    "node": ">= 14.17.0"
  },
```

Did I miss it in another place? @amurf

amurf:
Ah woops, must have overlooked that, just weary of it we been caught by it
before!

All good then 👍

package.json
```diff
@@ -40,15 +40,16 @@
     "//prepare": "prepare project for distribution",
     "prepare": "tsc",
     "//test": "runs tests on main project",
-    "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js"
+    "test": "tap --node-arg=-r --node-arg=ts-node/register --disable-coverage test/*_test.js",
+    "//test:cjs": "tests CommonJS compatibility (catches ESM-only deps)",
+    "test:cjs": "node --no-experimental-require-module scripts/test-cjs-compat.js"
```
miquelbeltran, line 45: Not a blocker, but this test command should be added to the ci job as well

miquelbeltran (approved):

